PR #23, opened 2013-06-25
Use simple mutexes for cases where RDLOCK is not native.

t3rmin4t0r:
This is exactly the same as using two mutexes because
for RDLOCK you do
  LOCK(lock->read);
for WRLOCK you do
  LOCK(lock->read);
the second lock on ->write is irrelevant because it exists within this mutex always.

So it is simpler & faster to use just a single lock.

Once that is in place, the RDLOCK for the native READLOCK has a race condition

ref_count++ | ref_count--

for the cache entry can run in two parallel processes with ambigous results.

The gcc/vc++ intrinsics for the atomic increments remove this race condition

__sync_add_and_fetch(ref_count) | __sync_sub_and_fetch(ref_count)

will run safely in parallel, as they lock each other out during the read-modify-write sections.


krakjoe:
I have implemented the simpler mutex for seven branch, I'm still looking at the other race condition thing ... I will eventually backport to 5 branch before doing a final release for 5 series ...

krakjoe:
The seven branch now has these, and hopefully some other problems solved ...

krakjoe:
@t3rmin4t0r omg ... I just realised what this was trying to fix ... I feel stupid ...

krakjoe:
I've genuinely fixed it now ... 

@t3rmin4t0r I just didn't realise I had omitted to put the atomics back in when I tore apart APC to make APCu ... a stupid stupid mistake, my fault ...

PR #92, opened 2014-10-24
fix undefined index error on user cache page

jensk:


lexvjatkin:
:+1:

Introduced here: https://github.com/krakjoe/apcu/commit/5e35ce1d71895c393b2d91cd93253410592164b1
but then fixed only one variable: https://github.com/krakjoe/apcu/commit/aba5bf3019cc7b522fa7627c0762d4a7c8a6ab40

screen8

PR #101, opened 2014-12-11
apc_cas: don't restrict old and new values to be integers

marc-mabe:
For now allow scalar values

The tests passed but I'm not sure about the memory allocation if the new value is of different size as the old one and I don't know if this PR is right to point to master.

### DIFF
--- a/php_apc.c
+++ b/php_apc.c
@@ -754,36 +754,68 @@ PHP_FUNCTION(apcu_dec) {
 
 /* {{{ php_cas_updater */
 static zend_bool php_cas_updater(apc_cache_t* cache, apc_cache_entry_t* entry, void* data) {
-    long* vals = ((long*)data);
-    long old = vals[0];
-    long new = vals[1];
+    zval** vals = ((zval**)data);
+    zval* old = vals[0];
+    zval* new = vals[1];
     zval* val = entry->val;
 
-    if (Z_TYPE_P(val) == IS_LONG) {
-        if (Z_LVAL_P(val) == old) {
-            Z_LVAL_P(val) = new;
+    // only scalars are allowed as new values
+    if (Z_TYPE_P(new) != IS_NULL
+        && Z_TYPE_P(new) != IS_BOOL
+        && Z_TYPE_P(new) != IS_LONG
+        && Z_TYPE_P(new) != IS_DOUBLE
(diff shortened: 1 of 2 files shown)


KOTRET:
what about the increased memory usage? any statistics comparing before / after?

marc-mabe:
I need to rebase this against the simplify branch to make it compile with PHP 5.6

Testscript:
```php
<?php

$iterations = 100000;
$s = $i = 0;
apcu_store('test', 0);

echo "apcu_cas INT -> INT: ";
$s = microtime(true);
for ($i = 0; $i < $iterations; ++$i) {
    apcu_store('test', 123);
    apcu_cas('test', 123, 456);
    apcu_cas('test', 456, 123);

    apcu_cas('test', 456, 123);
}
echo (microtime(true) - $s) . "\n";


echo "apcu_cas INT -> STR: ";
$s = microtime(true);
for ($i = 0; $i < $iterations; ++$i) {
    apcu_store('test', 123);
    apcu_cas('test', 123, 'test');
    apcu_cas('test', 'test', 123);

    apcu_cas('test', 'test', 123);
}
echo (microtime(true) - $s) . "\n";

echo "apcu_cas INT -> FLT: ";
$s = microtime(true);
for ($i = 0; $i < $iterations; ++$i) {
    apcu_store('test', 123);
    apcu_cas('test', 123, 456.789);
    apcu_cas('test', 456.789, 123);

    apcu_cas('test', 456, 123.456);
}
echo (microtime(true) - $s) . "\n";
```

With this change:
```
apcu_cas INT -> INT: 0.12929892539978
apcu_cas INT -> STR: 0.13439989089966
apcu_cas INT -> FLT: 0.1327919960022
```

Without this change (last two blocks commented out):
```
apcu_cas INT -> INT: 0.12774896621704
```

So there is a performance lost of ~1.2% for the function `apcu_cas` but it adds a vary basic feature and helps solving issues as before the arguments of `apcu_cas` where casted to int automatically.
```php
apcu_store('test', 123);
apcu_cas('test', 123, 456.789);
var_dump(apcu_fetch('test')); // = int(456)
```

krakjoe:
Feel free to rework your patch, it would be useful to provide a patch against seven branch also.

Going to close the PR, because wrong branch ...

PR #105, opened 2015-01-09
Fix apc.php UI code since vars from apcu_cache_info() have been renamed

sebastienbarre:
It looks like some variables returned by APCu have been renamed since the fork from APC.
To name a few:
- nhits was num_hits
- nmisses was num_misses
- ninserts was num_inserts
- nentries was num_entries
- nexpunges was num_expunges
- stime was start_time

Here is a fix for the apc.php UI to work.
Thanks

### DIFF
--- a/apc.php
+++ b/apc.php
@@ -375,11 +375,11 @@ if (isset($MYREQUEST['IMG']))
 		break;
 
 	case 2:
-		$s=$cache['num_hits']+$cache['num_misses'];
-		$a=$cache['num_hits'];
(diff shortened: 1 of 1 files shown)


sebastienbarre:
Hmm, not sure why my branch is off your 'simplify' branch, but seems it is even with 'master' as of now, it shouldn't be too bad. I can still fix that problem, if needed.

krakjoe:
This is fixed in the latest sources, please update your sources.

sebastienbarre:
My bad, I'm using binaries, I should have checked that tags. Thanks.

PR #107, opened 2015-02-16
fix APC compatibility

Stricted:
see https://github.com/krakjoe/apcu/commit/4f47d6dea51cf00ab13e623a6dd91c3d96c0fcc4#commitcomment-8126847


mishan:
:+1:

mishan:
@krakjoe

Stricted:
@krakjoe ...

mishan:
It'd be nice to keep keys compatible between APC and APCu so similar/same status pages can be used.

PR #124, opened 2015-08-27
Fix config.w32 for PHP7

Jan-E:
Hi Joe, I know it is work in progress, but if I want to follow the progress it might be welcome to fix config.w32 for the seven branch


krakjoe:
Thanks @Jan-E for taking the time ...

Jan-E:
No problem. You do the hard work. I am only testing and find things like this:
https://github.com/krakjoe/pthreads/issues/479

krakjoe:
Thanks again :)

PR #135, opened 2015-10-07
apcu_inc/dec should create missing cache entries

tony2001:
At the moment apcu_dec/inc() return FALSE when the entry they modify doesn't exist. This patch changes this behavior - they'll create missing entry and assign it 'step' value.


ksimka:
This has to be especially emphasized in docs after 'seven' branch is released. We implement this behaviour manually with add+inc, so it'll be BC break for us.

tony2001:
Why would that be a BC break for you?
If you add an empty value and then do an inc(), the behavior won't change a bit.

ksimka:
Oh, sorry, you're right, my bad.

PR #148, opened 2015-11-19
Fix error: Undefined index: key in line 1005

aguvillalba:
In APCu version 4.0.7 with PHP 5.5
apcu1
apcu1b


krakjoe:
We don't use the master branch, the branch for PHP 5 is "simplify", the branch for PHP 7 is "seven".

aguvillalba:
Ok thanks!

PR #149, opened 2015-11-19
Renamed index expunges to num_expunges and mtime to modification_time…

aguvillalba:
apcu1
apcu1b
apcu2
apcu2b

… to make it work with PHP 5.5 apcu v4.0.7


krakjoe:
apc.php at the head of the simplify branch is consistent with itself, this would break with the head of the branch.

A final release of apcu for five series will be done when the release for seven is done, it will be changing for the short form of these names, and staying that way ...

PR #190, opened 2016-07-24
Fix memory leak/failed check for duplicates when serializing arrays and object

TysonAndre:
Affects built in serializer as well as the igbinary7 serializer.

The apc serialization/unserialization code was checking
IS_REFCOUNTED_P(val).
However, when objects were being serialized as strings, the 
IS_REFCOUNTED bit in the type flags wasn't being set (Bitmask was 0)

This meant that if there were two uses of an object being serialized, it would be
cloned, since the code treated `dst` as if it couldn't be refcounted, and 
thus didn't add the mapping from `src`'s zend_refcounted pointer to `dst` into
`si->copied`

Found this bug while working on https://github.com/igbinary/igbinary7/pull/23

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -1234,7 +1234,10 @@ static zval* my_serialize_object(zval* dst, const zval* src, apc_context_t* ctxt
 		}
 
 		ZVAL_STR(dst, serial);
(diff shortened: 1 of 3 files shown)


TysonAndre:
@krakjoe @remicollet

PR #191, opened 2016-07-24
Print error message if an invalid zend_type is encountered when serializing/unserializing

TysonAndre:
Also, remove IS_CALLABLE. It's used for type hints.
It's not used anywhere for `zval` values.
E.g.

```c
Zend/zend_compile.c
4560:                           } else if (arg_info->type_hint ==
IS_CALLABLE && default_ast) {
```

assert() is replaced with a no-op in non-debug binaries.

If there is an unrecognized type, it is probably a symptom of buggy
extension, use of zval after free, or shared memory corruption.
It's useful to have an alert about that.

Instead of assert(0), return the value null and trigger an error.

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -1555,6 +1555,7 @@ static APC_HOTSPOT zval* my_copy_zval(zval* dst, const zval* src, apc_context_t*
 		}
 	}
 
+    /* Handle copying types types listed in https://wiki.php.net/phpng-int */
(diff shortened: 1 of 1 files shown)


krakjoe:
I'm not so sure about that ... whoever is debugging the buggy ext should be using a debug build, and so assert() is fine.

Happy to merge removal of IS_CALLABLE, well spotted ... open a new PR for that please.

PR #192, opened 2016-07-25
Fixes #188: Fix Segfault in ZTS build when locking

### LINKED ISSUE #188: Segfault on PHP 7.1
`composer self-update` crashes when apcu is enabled on php 7.1 (at least when done on travis-ci)

TysonAndre:
In ZTS builds with static TSRMLS,
blocking interruptions requires TSRMLS to be initialized.

(WLOCK will block signal interruptions, then acquire a lock)

The code initialized it in RINIT (per-request), but did not initialize
it in MINIT(module init).
But module init needed to lock in order to not be interrupted while
allocating an arena in shared memory.

Fixed this by calling ZEND_TSRMLS_CACHE_UPDATE();

Other extensions do this in MINIT in both php-7.0 and php-7.1
E.g.  https://github.com/php/php-src/blob/PHP-7.0.9/ext/mbstring/mbstring.c#L1550

This worked locally when tested on php configured with `--enable-maintainer-zts --enable-debug`

### DIFF
--- a/php_apc.c
+++ b/php_apc.c
@@ -210,6 +210,9 @@ static PHP_MINFO_FUNCTION(apcu)
 /* {{{ PHP_MINIT_FUNCTION(apcu) */
 static PHP_MINIT_FUNCTION(apcu)
 {
+#if defined(ZTS) && defined(COMPILE_DL_APCU)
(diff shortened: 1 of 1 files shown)


remicollet:
Indeed this fix the ZTS issue.

Merged

PR #193, opened 2016-08-10
Improve autotools AC_ARG_ENABLE usage

ghost:
This config.m4 falls into an autotools issue of present/not present vs enabled/disabled described at [https://autotools.io/autoconf/arguments.html](https://autotools.io/autoconf/arguments.html)

Previously, the macros appeared to have the use of "if enabled ; then x ; else y"

This is not how autoconf interprets the AC_ARG_ENABLE macro.
Instead it is "if present; then x; else y"

Now, it uses the result of the AC_ARG_ENABLE to test and act in a true IF.

The apparent desired behavior is preserved.

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -6,59 +6,52 @@ PHP_ARG_ENABLE(apcu, whether to enable APCu support,
 
 AC_MSG_CHECKING(if APCu should be allowed to use rwlocks)
 AC_ARG_ENABLE(apcu-rwlocks,
-[  --disable-apcu-rwlocks  Disable rwlocks in APCu],
-[
-  PHP_APCU_RWLOCKS=no
-  AC_MSG_RESULT(no)
-],
-[
-  PHP_APCU_RWLOCKS=yes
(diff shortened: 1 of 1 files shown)


krakjoe:
@remicollet do you have a build matrix with lots of different configurations by chance ?

remicollet:
I can test on RHEL-6 (the older version for which I build PHP 7)

So autoconf 2.63 - 2.69 => OK.

But perhaps we should prefer the PHP macros (PHP_ARG_ENABLE, PHP_ARG_WITH) instead of AC ones

krakjoe:
I feel unsure about why it used AC ones in the first place ... I'm scared of autotools ...

Anyone else got some input ?

madmajestro:
I'm closing this because it is obsolete since `AC_ARG_ENABLE` was replaced by `PHP_ARG_ENABLE` in f7fe4cc67d85c6004a6bc595182a3ff5183bdd62 / #517

PR #195, opened 2016-08-31
 Fixes #185 : Copy key's strings from smem into request mem

### LINKED ISSUE #185: zend_mm_heap corrupted
PHP-FPM segfaults with a log message: "zend_mm_heap corrupted".

I'm using versions 5.1.4 and 5.1.5 from remi repo (php70-php-pecl-apcu) on Centos 6.
 Both versions are segfaulting.
Version 5.1.3-1.el6.remi is OK and running for days in production without any problem.

$ uname -a
Linux myserver 2.6.32-504.30.3.el6.x86_64 #1 SMP 

Most of the time, we are using apc_inc() calls in our app.

TysonAndre:
Hopefully, this doesn't miss anything.
This fixes some memory leaks /invalid frees in tests/iterator_* when the tests
are run under valgrind.
    
This copies the keys from request mem into smem twice if you are doing
foreach($iterator as $key => $value), where $iterator has the flag
APC_ITER_KEY. I couldn't think of an simple way to de-duplicate, so
I didn't. (Would also have to handle possibly calling key() multiple times if
iterator was used manually)

Previously, the code was putting `zend_string`s with strings that were allocated as 

The test may be redundant.

### DIFF
--- a/apc_iterator.c
+++ b/apc_iterator.c
@@ -50,7 +50,9 @@ static apc_iterator_item_t* apc_iterator_item_ctor(apc_iterator_t *iterator, apc
 	}
 
 	if (APC_ITER_KEY & iterator->format) {
-		add_assoc_str(&item->value, "key", item->key);
+		/* item->key is allocated in shared memory. Return something which is emalloc()ed instead. */
+		zend_string *key_dup = zend_string_init(ZSTR_VAL(item->key), ZSTR_LEN(item->key), 0);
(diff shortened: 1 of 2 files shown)


krakjoe:
This was fixed by last commit ... thanks for taking the time to try though :)

remicollet:
Test added in 3c48da9de938cee96c5db15fd65e64010a587dfa

PR #198, opened 2016-09-25
Ensure that exiting process releases locks

mkoppanen:
Hey @krakjoe,

NOTE: not yet for merging

opening a pull request to work on getting robust mutexes to APCu. This is a first poke on implementing robustness (when a process exits it releases locks) to protect against hitting memory / time limit while holding lock (unsure if latter is prevent with the block_interruptions stuff).

Some implementation notes:

- pthread rwlocks don't support pthread_mutex_robust
- pthread mutex with robust mutex requires the mutex to be allocated to shared memory (based on my simple tests forking a mutex doesn't result in EOWNERDEAD if child exits while holding lock)

I was initially thinking about implementing simple rwlock using a mutex and cond variables but probably native platform primitives provide better performance. Although the current implementation is a bit nasty, it's a starting point for discussion on robust mutexes.

More reading: http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutex_consistent.html

### DIFF
--- a/apc_lock.c
+++ b/apc_lock.c
@@ -22,6 +22,107 @@
 # include "apc_lock.h"
 #endif
 
+#ifndef PHP_WIN32
+# ifndef APC_SPIN_LOCK
+#   ifndef APC_FCNTL_LOCK
+#       ifdef APC_LOCK_RECURSIVE
+#           define APC_LOCK_ROBUST 1
+#       endif
+#   endif
+# endif
+#endif
+
+
+
+#ifdef APC_LOCK_ROBUST
+
+#include "apc_stack.h"
+
+/*
+	rwlocks don't support robustness so we have to hack around it
+*/
+static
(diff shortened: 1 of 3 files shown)


krakjoe:
This is causing many problems ...

I've read the patch, it's not so nasty ... if it works ...

@mkoppanen current state of this patch ?

@weltling what has windows got here ?

mkoppanen:
@krakjoe,

I think in terms of cleanup supporting native robust mutexes would be ideal (http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutex_consistent.html) as an addition. Other than that I am not sure if the assumptions made here are correct (unpaired locks can be costlier etc).

From what I can see there is no way to safely call any PHP API function that might call e* family allocators because being close to memory limit could trigger hard exit. thats quite 👎 from callers perspective.

PR #219, opened 2016-12-23
fix zend_alter_ini_entry() which segfault under cli.

userid:
when running php5-fpm already with apcu.ini:
```
extension=apcu.so
apc.enabled = 1
;apc.enable_cli = 1
```
then, execute php-cli:
```
$php -r ' var_dump("a"); '
string(1) "a"
Segmentation fault
```
however if add `apc.enable_cli = 1` to `apcu.ini`, then it works.
```
$php -r ' var_dump("a"); '
string(1) "a"
```
no Segmentation fault now.

it happened in both php5.4 and php5.6.


nikic:
The PHP 5 version of apcu is no longer supported and this code does not seem to be present in the PHP 7 version, so closing here.

Issue #223, opened 2017-01-19
Fatal error when starting php on windows


madhon:
C:\php>php --version
PHP Fatal error:  PHP Startup: apc_shm_attach: shmat failed: in Unknown on line
0

The ini file just has the following
;extension=php_apcu.dll
apc.enabled=1
apc.shm_size=32M
apc.ttl=7200
apc.enable_cli=1

This is with PHP 7.0.14 NTS x86 VC14 on Windows 2012

Cy4n1d3:
Happens for me aswell, PHP 7.0.16 NTS x64 on Windows 2008 R2.

Fludimir:
same fo me with PHP 7.1.1 TS x64 on Windows7 x64

Jan-E:
Try loading php_sysvshm.dll before php_apcu.dll:

```
extension=php_sysvshm.dll
extension=php_apcu.dll

```

Cy4n1d3:
Thanks for your input Jan-E, sadly loading `php_sysvshm.dll` still results in this error:

    [20-Feb-2017 12:31:41 Europe/Berlin] PHP Fatal error:  PHP Startup: apc_shm_attach: shmat failed: in Unknown on line 0

Jan-E:
Bugger. Try my builds of php_sysvshm.dll and php_apc.dll:
https://www.apachelounge.com/viewtopic.php?t=6359

I know for sure that they load fine. See for instance
https://phpdev.toolsforresearch.com/php-7.0.16-nts-Win32-VC14-x64.htm

That phpinfo() dump is created by running 'php-cgi.exe phpinfo.php' on the command line with the php.ini that is stored as php.ini.bak now in https://phpdev.toolsforresearch.com/php-7.0.16-nts-Win32-VC14-x64.zip

If you try to reproduce that exactly, you will probably miss some supporting dll's for extensions like oci, interbase, firebird, postgress, imagick and the like. But you'll get the idea, I guess.

Jan-E:
BTW for @krakjoe I stumbled into this issue, because my hoster moved our production Win2008 R2 VPS to a new platform and I got exactly the same error with PHP 5.6.30 x86 NTS and apcu 4.0.11.

After the initial attach error the subsequent ones were like:

[...]

PR #244, opened 2017-04-14
Issue223

Jan-E:
I know it is a nasty workaround, but writing to the error log before we initialize the SHM seems to fix https://github.com/krakjoe/apcu/issues/223


nikic:
I believe this has already been fixed by https://github.com/krakjoe/apcu/commit/b3c53e89a37f98ca4edf6eaa9e8e18f87b9a9c74, so closing this.

Issue #247, opened 2017-06-05
when a NUL char is used as key, apcu_fetch(array) truncates the key


nicolas-grekas:
```php
apcu_store(array("a\0b" => 'foo'));
print_r(apcu_fetch(array("a\0b")));
```
output:
```
Array
(
    [a] => foo
)
```
expected:
```
Array
(
    [a<non-displayed-nul>b] => foo
)
```

7snovic:
I think that the expected result here is : `[a\0b] => foo` while as I see that the actual cache index is `a\0b`

nicolas-grekas:
@7snovic nowhere else are values "addlshashed"
this would be inconsistent with the way non-array calls behave (`apcu_store("a\0b", 'foo')`).

7snovic:
@nicolas-grekas I will try to give an explanation for this ASAP.

PR #252, opened 2017-07-29
Fix 247

7snovic:
a quick fix for #247 , @krakjoe


nicolas-grekas:
nowhere else are values "addlshashed"
this would be inconsistent with the way non-array calls behave.

nikic:
I've implemented a different fix in #272. As @nicolas-grekas said, values should never be implicitly addslashed.

PR #253, opened 2017-08-02
APCIterator results should include cache entries considered stale due to apc.ttl

davidsickmiller:
Depending on the apc.ttl setting, apc_fetch() "sees" data that APCIterator doesn't.  This seems incorrect to me.  I believe apc_fetch() is working correctly, so this PR make APCIterator work the same.


nikic:
This issue has been resolved as part of general TTL handling cleanup in #291 (part of apcu 5.1.12). Sorry for not using this PR, the changes here were all right.

Issue #266, opened 2017-11-17
Bdragon's APCU lock debugging adventure


bdragon28:
When PHP is compiled without Zend Signal Handling enabled, and pthread_rwlock_wrlock et al are in use, the generic locking macros in apc_lock_api.h are unstable and can get into priority inversion situations or get stuck locks, becasue HANDLE_BLOCK_INTERRUPTIONS() and HANDLE_UNBLOCK_INTERRUPTIONS() do not do anything without zend signal handling enabled, so the critical sections end up unprotected!

Currently, debian packages are being built without zend signal handling enabled. I'm unsure why, I examined a recent buildd log and it doesn't look like it's explicitly disabling it in the configure options, yet it's showing "checking whether to enable zend signal handling... no". If I manage to track down why, I'm going to push for getting it enabled in the future.

To protect things in general, I think it would probably be best to force PHP_APCU_RWLOCKS=no when ZEND_SIGNALS is not defined. Although I really don't know if this is sufficient, I'm not an expert at locking.

I suspect this is the root cause for the hangs in pthread_rwlock_wrlock() that have been cropping up lately.

bdragon28:
Oh, of course, the default enable for ZEND_SIGNALS was only added in PHP 7.1! So it would have to be explictly enabled on 7.0.

There *used* to be a fallback where the sapi could provide the functions instead, but it got disabled in https://github.com/php/php-src/commit/127db5a37d178caac24f4eabbc6695923864e60a and was later removed.

This would explain why PHP 7.0 packages tend to have nonfunctional HANDLE_BLOCK_INTERRUPTIONS() / HANDLE_UNBLOCK_INTERRUPTIONS() macros -- package maintainers need to configure with --enable-zend-signals because the default of not enabling them leaves them broken!

bdragon28:
I take back the priority inversion comment, that's not a concern. Locks hanging around forever is the concern. I guess the main reason of these checks is to make the critical sections crash resistant. I *have* been seeing SIGBUS errors in operation, and that seems like the sort of thing that would tend to happen exactly where the problems have been occurring (smack dab in apc_cache_insert during the php_apc_try().... I guess it probably dies before being able to run the bailout.)

OK, I see that you do explicitly install a SIGBUS handler. Not convinced that apc_core_unmap is always running properly, the lock is definitely getting stuck.

Will bang on it with a larger hammer.

bdragon28:
[...]

PR #267, opened 2017-11-22
Stuff

nikic:
* Compatibility with PCRE2 in PHP 7.3
* Set IS_STR_PERSISTENT so refcounting failures are reported by `-DRC_DEBUG=1`
* Fix refcounting errors in APCIterator. This coincides with the issues found in #266.

### DIFF
--- a/apc_iterator.c
+++ b/apc_iterator.c
@@ -50,7 +50,7 @@ static apc_iterator_item_t* apc_iterator_item_ctor(apc_iterator_t *iterator, apc
 	}
 
 	if (APC_ITER_KEY & iterator->format) {
-		add_assoc_str(&item->value, "key", zend_string_copy(item->key));
+		add_assoc_str(&item->value, "key", zend_string_dup(item->key, 0));
 	}
 
     if (APC_ITER_VALUE & iterator->format) {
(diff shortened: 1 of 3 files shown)


nikic:
This breaks CI because Travis is stuck on a very old nightly.

apc_iterator.c
```diff
@@ -325,14 +332,20 @@ void apc_iterator_obj_init(apc_iterator_t *iterator, zval *search, zend_long for
     iterator->regex = NULL;
     iterator->search_hash = NULL;
     if (search && Z_TYPE_P(search) == IS_STRING && Z_STRLEN_P(search)) {
+		uint32_t capture_count;
 #ifdef ITERATOR_PCRE
         iterator->regex = zend_string_copy(Z_STR_P(search));
-        iterator->re = pcre_get_compiled_regex(iterator->regex, NULL, NULL);
+        iterator->re = pcre_get_compiled_regex(iterator->regex, &capture_count, NULL);
```
bdragon28: pcre1 equivilent for old php is pcre_extra* in that position

nikic: Ah, right. I've just switched this to directly use `pcre2_match_data_create_from_pattern` instead, as I don't think it makes sense to hog the global mdata for this use-case anyway, as the iterator is a bit longer-lived.

bdragon28:
Working great against php 7.0.25. I've been hammering it locally and am no longer able to get iterators to crash, both with ZEND_SIGNALS enabled and without.

krakjoe:
Thanks for making the time nikita

staabm:
@krakjoe awesome thx. could we get a new release?

staabm:
@krakjoe friendly ping.. would it make sense to get a new release as this changes landed on master?

krakjoe:
Release requested, @remicollet will run QA first

remicollet:
@krakjoe can you please check/review changelog (package.xml) before 5.1.9 release ?

Issue #295, opened 2018-03-12
ApcuIterator->rewind is slow


staabm:
in one of my websites I can see ApcuIterator->rewind() beeing the most expensive method call in the whole page.

![grafik](https://user-images.githubusercontent.com/120441/37274226-565426a2-25dc-11e8-8261-63e1d44d889b.png)

the call happens in the following spot (my app itself and the framework we use is closed source though)
https://github.com/staabm/thincache/blob/6585955bab9ccbc0b495462d7219776b336a6d5f/src/thincache/CacheApcu.php#L105-L121

if you need more information please feel free to ask.

env
```
This program makes use of the Zend Scripting Language Engine:
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.27-1+ubuntu16.04.1+deb.sury.org+1, Copyright (c) 1999-2017, by Zend Technologies
    with blackfire v1.18.1~linux-x64-non_zts70, https://blackfire.io, by SensioLabs

APCu Version | 5.1.9
```

nikic:
Assuming there are less than `$limit` entries in the result set, all the items will be fetched in `rewind()`. This involves iterating the *entire* cache and matching every entry, while under read-lock. If you have a large cache, I'd expect this to be quite slow.

staabm:
just fyi. we can do the same stuff using the CacheMemcached adapter which is 15% faster

https://github.com/staabm/thincache/blob/6585955bab9ccbc0b495462d7219776b336a6d5f/src/thincache/CacheMemcached.php#L120-L170

not sure whether memcache is more efficient for this kind of operation or its just a difference of different storage size / cached-keys count.

nikic:
@staabm A difference might be that apc currently is not making use of the pcre jit. That should be a relatively simple addition that might give a nice win here.

nikic:
I've implemented two optimizations in #297: Using the PCRE JIT, which should help your case (I'm assuming many cache entries, few matching results) and preallocating the string keys (which helps in the case where are many results). For the second point, you can already try if using `APC_ITER_KEY|APC_ITER_VALUE` (or whatever parts you actually need) instead of `APC_ITER_ALL` will give you a performance improvement.

Based on a microbenchmark I'm seeing >1.5x speedup, though it will depend on the specific usecase.

staabm:
[...]

PR #297, opened 2018-03-12
Improve APC iterator performance

nikic:
* Preallocate all the used strings, so they are not allocated anew for every single entry.
 * Allow the PCRE JIT to be used (only relevant for PHP <= 7.2, on master it was already used) for the search regex.


staabm:
cool thx. did you measure how much this changes will improve the iterator? if so, would be great if you could share this information.

if not - nevermind.

staabm:
my question about the benchmark was already answered in https://github.com/krakjoe/apcu/issues/295#issuecomment-372482845

PR #305, opened 2018-03-17
Implement precisely-allocated persistence

nikic:
Currently, persistence to SMA is pool based, i.e. SMA memory is allocated in blocks (of minimum size 512 bytes). This means that small entries (say entries storing integers) use much more memory than they need to and large entries will have a variable amount of additional overhead (depending on how many pool blocks are needed, how well they are filled and which block sizes are used).

This PR implementations a new mechanism, which performs persistence in two steps: First, we compute the amount of memory the persistence will require, then we perform a single SMA allocation of the correct size, and finally perform the actual copies into the allocated memory segment.

As such, this implementation only performs a single allocation of exactly the right size, at the cost of having to precompute the size in a separate pass. We also save 64 bytes in pool management overhead as a nice side benefit.

The implementation is not quite ready yet, some things left:

 * [x] This still uses the old code from copying from SHM. This can be simplified a lot now.
 * [x] Missing handling for INDIRECT zvals in arrays (can only occur in the unusal case of persisting the symbol table). There's currently no test for this.
 * [x] Missing test for nNextFreeElement handling. The previous implementation was buggy in this regard.
 * [x] Performance testing.


nikic:
Some microbenchmark numbers using https://gist.github.com/nikic/7b1abdc9a43c17ed20a635ac6f93f441:

```
OLD (default serializer):

int store:      0.05s, 58M
int fetch:      0.25s, 58M
string store:   0.57s, 565M
string fetch:   1.01s, 565M
array store:    0.79s, 2169M
array fetch:    2.92s, 2169M
object store:   0.05s, 58M
object fetch:   0.58s, 58M

NEW (default serializer):

int store:      0.03s, 16M
int fetch:      0.16s, 16M
string store:   0.56s, 496M
string fetch:   0.93s, 496M
array store:    0.83s, 526M
array fetch:    1.32s, 526M
object store:   0.04s, 21M
object fetch:   0.44s, 21M

OLD (php serializer):

int store:      0.05s, 58M
int fetch:      0.25s, 58M
string store:   0.58s, 565M
string fetch:   1.03s, 565M
array store:    1.16s, 460M
array fetch:    14.90s, 460M
object store:   0.05s, 58M
object fetch:   0.59s, 58M

NEW (php serializer):

int store:      0.03s, 16M
int fetch:      0.17s, 16M
string store:   0.56s, 496M
string fetch:   0.95s, 496M
array store:    1.16s, 392M
array fetch:    14.33s, 392M
object store:   0.04s, 21M
object fetch:   0.44s, 21M
```

staabm:
Just reading the numbers: it seems storing a array into apcu is waaaay slower then any other type.
Should we try to get arround storing a array from a userland perspective?

nikic:
Due to a benchmark error I did not notice the regression in array persist/unpersist performance. With #324 the numbers there look better:

```
int store:      0.03s, 16M
int fetch:      0.17s, 16M
string store:   0.58s, 496M
string fetch:   0.96s, 496M
array store:    0.94s, 1549M
array fetch:    1.74s, 1549M
object store:   0.04s, 21M
object fetch:   0.46s, 21M
```

@staabm The benchmark script compares very different cases. "array" are arrays with sizes between 0 and 1000, while the objects are always empty `stdClass`es. Generally objects are the slowest type to store/fetch, because they will always use PHPs serialization mechanism (which is extremely slow). This is not intended as a comparison between different types.

PR #308, opened 2018-03-26
Keep search parameter on cache entry detail link.

DennisBirkholz:
Detail/entry content then works for cache entries that are not on the first page of the unfiltered list.

### DIFF
--- a/apc.php
+++ b/apc.php
@@ -64,7 +64,7 @@ $time = time();
 $host = php_uname('n');
 if($host) { $host = '('.$host.')'; }
 if (isset($_SERVER['SERVER_ADDR'])) {
-  $host .= ' ('.$_SERVER['SERVER_ADDR'].')';
+	$host .= ' ('.$_SERVER['SERVER_ADDR'].')';
 }
 
 // operation constants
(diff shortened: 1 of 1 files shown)


DennisBirkholz:
I renamed the newly introduced variable to better show its meaning.
I also fixed indentation/whitespace with a separate commit.

As my changes (or the addresses problem) did not seem to explain itself, a short description of the problem:

Lets assume you have 26 entries in your cache: `entry_a` to `entry_z`. If you now search for `z` only `entry_z` is displayed on the result page. If you now click on `entry_z` to view its contents, your search is discarded and you are back to a list view. If `entry_z` happens to be in the entries of that view, its contents is displayed, otherwise not.

My patch now keeps the search string in the link to display the entry detail so if you click on an entry, you see the same search result list but with the clicked entry expanded.

nikic:
Sounds reasonable, but I won't be able to test and merge this until next week.

DennisBirkholz:
> I won't be able to test and merge this until next week.

No problem, have a nice easter holiday ;-)

nikic:
Thanks!

PR #318, opened 2018-08-17
Testing fixes

DaveRandom:
General cleanup of `server_test.inc`, including better handling of `PHP_TEST_ARGS` environment variable. This variable is used to pass general arguments to `run-tests.php`, so rather than passing it verbatim to PHP for child processes it should be parsed to extract ini settings passed via `-d` and only these values should be forwarded.

Also update the bc submodule to include the [fixes to the tests](https://github.com/krakjoe/apcu-bc/pull/26) that have been merged there.

### DIFF
--- a/tests/apc54_014.phpt
+++ b/tests/apc54_014.phpt
@@ -29,7 +29,7 @@ $args = array(
 	'apc.preload_path=' . dirname(__FILE__) . '/data',
 );
 
-$num_servers = 1;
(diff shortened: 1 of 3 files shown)


nikic:
server_test.inc probably comes from php-src upstream. Does this sync with upstream changes or fork the implementation?

DaveRandom:
Closed in favour of #319

PR #326, opened 2018-10-09
Improve mmap error handling

nikic:
* Use `E_CORE_ERROR` to make sure startup is aborted -- otherwise we would segfault lateron.
* Print a better error message, including the size of the allocation.

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -286,11 +286,11 @@ PHP_APCU_API apc_cache_t* apc_cache_create(apc_sma_t* sma, apc_serializer_t* ser
 	cache->shmaddr = apc_sma_malloc(sma, cache_size);
 
 	if (!cache->shmaddr) {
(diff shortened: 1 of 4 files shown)


apc_cache.c
```diff
@@ -285,8 +285,8 @@ PHP_APCU_API apc_cache_t* apc_cache_create(apc_sma_t* sma, apc_serializer_t* ser
 	/* allocate shm */
 	cache->shmaddr = apc_sma_malloc(sma, cache_size);
 
-	if(!cache->shmaddr) {
-		apc_error("Unable to allocate shared memory for cache structures.  (Perhaps your shared memory size isn't large enough?). ");
+	if (!cache->shmaddr) {
+		zend_error_noreturn(E_CORE_ERROR, "Unable to allocate shared memory for cache structures. Perhaps your shared memory size isn't large enough?");
```
staabm: should this error message contain the `cache_size`?

nikic: Done in https://github.com/krakjoe/apcu/pull/326/commits/484ded0174546c02df1f68a07ad18122d9a50f62.

PR #334, opened 2018-11-16
Make apcu_cas require only read-lock

nikic:
This changes the implementation of apcu_cas to only acquire a read-lock on the cache (to prevent a concurrent write), and perform the update itself using atomic compare-and-swap.

I've not updated inc/dec here, because I'm unsure on how to handle the overflow behavior. Right now apcu_inc/dec will overflow into double (at which point it will stop being incremented). If we switch this to atomic we'd get a wraparound instead.

I'm not sure how worthwhile this change is overall.

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -898,6 +898,53 @@ retry_update:
 }
 /* }}} */
 
+/* {{{ apc_cache_atomic_update_long */
+PHP_APCU_API zend_bool apc_cache_atomic_update_long(
+		apc_cache_t *cache, zend_string *key, apc_cache_atomic_updater_t updater, void *data,
+		zend_bool insert_if_not_found, zend_long ttl)
+{
+	apc_cache_entry_t *entry;
+	zend_bool retval = 0;
+	time_t t = apc_time();
+
+	if (!cache) {
+		return 0;
+	}
+
+retry_update:
+	APC_RLOCK(cache->header);
+	entry = apc_cache_rlocked_find_nostat(cache, key, t);
+	if (entry) {
+		/* Only supports integers */
+		if (Z_TYPE(entry->val) == IS_LONG) {
+			retval = updater(cache, &Z_LVAL(entry->val), data);
+			entry->mtime = t;
+		}
+
+		APC_RUNLOCK(cache->header);
+		return retval;
+	}
+
+	APC_RUNLOCK(cache->header);
+	if (insert_if_not_found) {
+		/* Failed to find matching entry. Add key with value 0 and run the updater again. */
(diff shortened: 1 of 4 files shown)


apc_cache.c
```diff
@@ -898,6 +898,53 @@ PHP_APCU_API zend_bool apc_cache_update(
...
+	zend_bool retval = 0;
+	time_t t = apc_time();
+
+	if (!cache) {
+		return 0;
+	}
+
+retry_update:
+	APC_RLOCK(cache->header);
+	entry = apc_cache_rlocked_find_nostat(cache, key, t);
```
LionsAd: As the updater is quick and there is just one ATOMIC operation, it is not faster to get an incref lock on the entry itself. (as that would be 3 atomic operations).

But I wondered:

Would it not be enough to have an incref on the entry to be able to update it in general?

---

I'll do some benchmarks with this PR :)

nikic: The incref part is what allows you to use the entry after releasing the rlock -- but this only makes sense if you expect to perform some expensive operation on it. The updater is cheap enough to call it while still holding the rlock.

Acquiring the rlock is necessary to avoid concurrent modifications while the list of entries is traversed. Supporting that without a lock would require a lock-free table.

LionsAd: Yep, that is what I meant: If in general instead of insert + delete for apcu_store() it could be enough to do an update with increased incref.

LionsAd:
Some more benchmarks for only `cas` call:

Before: 4.151 seconds
After: 2.555 seconds

=> 60% faster

This really is worth it.

LionsAd:
I think we should update _inc / _dec as well -- even if they wrap-around. (but that can be in a follow-up)

I'd like to use that to more intelligently determine when to store data in APCu and when it is not accessed enough to be worthwhile - for that I need inc() to be fast ...

nikic:
> I'd like to use that to more intelligently determine when to store data in APCu and when it is not accessed enough to be worthwhile - for that I need inc() to be fast ...

For that purpose, maybe using `apcu_key_info()` would be useful? It includes a hit count.

LionsAd:
@nikic Ohhh, yes indeed. apcu_key_info() is right now undocumented on php.net, which is why I did not see it. That is even better as I can just do a:

`apcu_add();`

on potential candidates

and then `apcu_fetch()` + `apcu_key_info()` whenever they are accessed.

LionsAd:
@nikic Can we merge that now or do you still have concerns?

nikic:
Merged! Thanks for performing benchmarks!

Issue #335, opened 2018-11-17
APCu stampede protection is broken


LionsAd:
Splitting out as a separate bug from https://github.com/krakjoe/apcu/issues/333#issuecomment-439498081 so I can work on it

Issue #333, opened 2018-11-14
apcu_store locking is breaking down under high concurrency


LionsAd:
Something still is not quite right with apcu_store locking (5.1.11):

I saw up to 6 seconds for apcu_store() calls in production.

With my test script I can have it at least take 3.5 ms per call instead of 0.08 ms per call.

And _sometimes_ I can lock it after running high concurrency load (but happened just once and I did not get a trace).

With that script I can see that there are apcu_fetch / apcu_store sequences taking 300-445 ms per call -- which should never happen.

Test script is here:

```
<?php

$item = (object) [
  'data' => 'x',
];

for ($i = 0; $i < 1000; $i++) {
$cid = 'fabian:' . mt_rand(0,20);

$item = (object) [
  'created' => time(),
  'cid' => $cid,
  'data' => [
     'foo' => 'bar',
     'x' => $item->data,
  ],
];

$start = microtime(TRUE);

$old_item = apcu_fetch($cid);
apcu_store($cid, $item);

$diff = (microtime(TRUE) - $start) * 1000;

if ($diff > 10) {
  error_log(print_r([$diff], TRUE), 3, '/tmp/apcu.log');
}

echo $cid . PHP_EOL;
}
```

It only breaks down under high concurrency, with low concurrency the script takes 200 ms, with high concurrency, it takes up to 6 seconds and both _fetch and _store are slower then.

To test I used `ab -n10000 -c200 http://127.0.0.1/test.php`

Edit: The script is useless unfortunately as ab aborts the request once it sees the `echo` and the 1000 loop is too much data in the array. Trying to find a better example.

nikic:
I think that most likely what you're hitting here is not really an implementation bug, but just the fact that apcu has (essentially) a single global rwlock for the whole cache. Every read and write has to acquire a cache lock. While there can be multiple concurrent reads, writes are exclusive. If you have high concurrency and a write-heavy workload, there's going to be a lot of lock contention and some of the processes will have to wait (and potentially will be descheduled, while will drive up pause times by a lot).

The solution to this would be to introduce finer grained locking (or lock-free, but let's take this one step at a time...). The obvious way to do that would be to have a separate lock for each hashtable bucket, which would significantly reduce lock contention for the (hopefully common) case where accesses are distributed across hashtable buckets.

The other problem you may be hitting is that if SHM runs full, a store might trigger a full cache cleanup. This will lock the cache and then proceed to remove all entries, which may take a considerable amount of time.

[...]

PR #336, opened 2018-11-17
Issue #335: Fix stampede protection

LionsAd:
As suggested in https://github.com/krakjoe/apcu/issues/333#issuecomment-439498081 this is the better fix.

Fixes: #335


apc_cache.c
```diff
@@ -1115,44 +1115,39 @@ PHP_APCU_API void apc_cache_stat(apc_cache_t *cache, zend_string *key, zval *sta
...
-		/* check the hash and length match */
-		if (last->hash == ZSTR_HASH(key) && last->len == ZSTR_LEN(key)) {
-			apc_cache_owner_t owner;
+		apc_cache_owner_t owner;
 #ifdef ZTS
-			owner = TSRMLS_CACHE;
+		owner = TSRMLS_CACHE;
 #else
-			owner = getpid();
+		owner = getpid();
```
nikic: This will now always perform a getpid() call, even if hash/len do not match. It's probably not particularly expensive due to vDSO, but maybe best to avoid it?

tests/ghbug335.phpt
```diff
@@ -0,0 +1,30 @@
+--TEST--
+GH Bug #335: APCu stampede protection is broken
+--SKIPIF--
+<?php require_once(dirname(__FILE__) . '/skipif.inc'); ?>
+--INI--
+apc.enabled=1
+apc.enable_cli=1
+apc.serializer=default
```
nikic: This line shouldn't be necessary, as this test does not rely on any particular serializer. Leaving out the option here will make sure that CI tests both default and php serializers.

tests/ghbug335.phpt
```diff
@@ -0,0 +1,30 @@
+--TEST--
+GH Bug #335: APCu stampede protection is broken
+--SKIPIF--
+<?php require_once(dirname(__FILE__) . '/skipif.inc'); ?>
```
nikic: This also needs something like `if (!extension_loaded('pcntl') die('skip pcntl required');`, as pcntl is an optional extension (and iirc not available on Windows).

apc_cache.c
LionsAd: But we need to update the owner in the `last` entry or not?

tests/ghbug335-fail.phpt
```diff
@@ -0,0 +1,34 @@
+--TEST--
+GH Bug #335: Test that APCu stampede protection can be disabled
+--SKIPIF--
+<?php require_once(dirname(__FILE__) . '/skipif.inc'); ?>
+if (!extension_loaded('pcntl') die('skip pcntl required');
```
nikic: This line should be part of the `<?php ?>` block :)

LionsAd: Uhhhh, indeed ...

nikic:
@LionsAd I'd just drop the message. As nobody is going to run a debug build in production, nobody will see those messages anyway...

nikic:
Okay, looks like Travis is using a ZTS build of PHP. I can reproduce the failure locally under `--enable-maintainer-zts`.

On ZTS the owner is determined by `TSRMLS_CACHE`, but if we fork that's not going to change. I wonder if that only happens in this particular case (fork in PHP script) or if this also doesn't work in fpm.

LionsAd:
Looks great now, thanks for fixing that! :)

nikic:
Squashed and merged as https://github.com/krakjoe/apcu/commit/c4358ad81e06ead790431f5392eab85693c2e8aa. Thanks for working on this!

Issue #338, opened 2018-11-17
Use mutex for SMA lock


nikic:
Currently the lock for the shared memory allocator also uses an rwlock. However rlocks are only used for getting allocator statistics, which is not a common operation. As such, this lock is almost entirely used for write locks. It would be more efficient to directly use a mutex instead.

The main part of making this work would be to change the lock abstraction to support both rwlocks and mutexes at the same time, rather than just one of them.

LionsAd:
@nikic Can we also remove all the cruft from the apc_lock.c while on it? It feels very hard to work with right now ...

I would suggest we have:

- RLOCK, WLOCK / RUNLOCK / WUNLOCK
- MLOCK / MUNLOCK

for mutex locks vs. RWLOCKS

LionsAd:
I have that implemented, benchmarks for `ab -n80 -c40` with a 50000 write loop of 21 different keys:

Before: 8.795 seconds
After: 7.336 seconds

So 20% faster out of the box.

Right now the main wlock is the bottleneck however, but with that removed speed near lockless is achieved:

- Lockless write / read lock + lockless SMA lock: 2.2 seconds
- Lockless write / read lock + SMA mutex lock: 2.852 seconds

Well worth it.

LionsAd:
It can well be that changing the write lock for the SMA to a mutex will fix the remaining deadlocks.

Currently I am able to consistently make APCu slow / deadlock (?) when removing all locks from write / read hashtable and use rwlocks for SMA ...

LionsAd:
But with mutex() that never happens.

nikic:
> It can well be that changing the write lock for the SMA to a mutex will fix the remaining deadlocks.
> 
> Currently I am able to consistently make APCu slow / deadlock (?) when removing all locks from write / read hashtable and use rwlocks for SMA ...

That seems very weird to me. If the rwlock is used exclusively for write-locks I would expect that rwlock and mutex (apart from perf characteristics) will work the same way. The only difference here I can think of is that the mutex is constructed as recursive, while rwlocks are not write-recursive (only read-recursive). However I don't see any place in the SMA code where we could lock recursively.

LionsAd:
I think I see a problem:

```
                        SMA_UNLOCK(sma, i);
                        sma->last = i;
```

This means that if someone gets a LOCK in on sma->last afterwards it would free a different lock then instead ...

Before getting a lock we need to save sma->last, so that locks are consistently locked and freed.

[...]

PR #339, opened 2018-11-18
Issue 338: Use mutex for SMA locks

LionsAd:
This has three parts:

- a) Add a mutex implementation
- b) Abstract out the lock implementation in SMA
- c) Use the mutex implementation in SMA

As written already in #338 benchmarks are favorable (20% faster):

Before: 8.795 seconds
After: 7.336 seconds

With less concurrency and requests (`ab -n40 -c5`):

Before: 4.8 seconds
After: 3.7 seconds
=> 29% faster

Fixes: #338

### DIFF
--- a/apc_mutex.c
+++ b/apc_mutex.c
@@ -1,0 +1,77 @@
+/*
+  +----------------------------------------------------------------------+
+  | APCu                                                                 |
+  +----------------------------------------------------------------------+
+  | Copyright (c) 2013 The PHP Group                                |
+  +----------------------------------------------------------------------+
+  | This source file is subject to version 3.01 of the PHP license,      |
+  | that is bundled with this package in the file LICENSE, and is        |
+  | available through the world-wide-web at the following url:           |
+  | http://www.php.net/license/3_01.txt                                  |
+  | If you did not receive a copy of the PHP license and are unable to   |
+  | obtain it through the world-wide-web, please send a note to          |
+  | license@php.net so we can mail you a copy immediately.               |
+  +----------------------------------------------------------------------+
+  | Author: Fabian Franz <fabian@lionsad.de>                             |
+  +----------------------------------------------------------------------+
+ */
+#include "apc_mutex.h"
+
+#ifdef APC_HAS_PTHREAD_MUTEX
+
+static zend_bool apc_mutex_ready = 0;
+static pthread_mutexattr_t apc_mutex_attr;
+
+PHP_APCU_API zend_bool apc_mutex_init() {
+	if (apc_mutex_ready) {
+		return 1;
+	}
+	apc_mutex_ready = 1;
+
+	if (pthread_mutexattr_init(&apc_mutex_attr) != SUCCESS) {
+		return 0;
+	}
+
+	if (pthread_mutexattr_setpshared(&apc_mutex_attr, PTHREAD_PROCESS_SHARED) != SUCCESS) {
+		return 0;
+	}
+
+	return 1;
+}
+
+PHP_APCU_API void apc_mutex_cleanup() {
+	if (!apc_mutex_ready) {
+		return;
+	}
+	apc_mutex_ready = 0;
+
+	pthread_mutexattr_destroy(&apc_mutex_attr);
+}
+
+PHP_APCU_API zend_bool apc_mutex_create(apc_mutex_t *lock) {
+	pthread_mutex_init(lock, &apc_mutex_attr);
+	return 1;
+}
+
+PHP_APCU_API zend_bool apc_mutex_lock(apc_mutex_t *lock) {
+	HANDLE_BLOCK_INTERRUPTIONS();
+	if (pthread_mutex_lock(lock) == 0) {
+		return 1;
+	}
+
+	HANDLE_UNBLOCK_INTERRUPTIONS();
+	apc_warning("Failed to acquire lock");
+	return 0;
+}
+
+PHP_APCU_API zend_bool apc_mutex_unlock(apc_mutex_t *lock) {
+	pthread_mutex_unlock(lock);
+	HANDLE_UNBLOCK_INTERRUPTIONS();
+	return 1;
+}
+
+PHP_APCU_API void apc_mutex_destroy(apc_mutex_t *lock) {
+	pthread_mutex_destroy(lock);
+}
+
+#endif
--- a/apc_mutex.h
+++ b/apc_mutex.h
@@ -1,0 +1,63 @@
+/*
+  +----------------------------------------------------------------------+
+  | APCu                                                                 |
+  +----------------------------------------------------------------------+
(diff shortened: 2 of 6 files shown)


apc_lock.c
```diff
@@ -346,3 +346,4 @@ PHP_APCU_API void apc_lock_destroy(apc_lock_t *lock) {
 }
 #endif
 
+#include "apc_mutex.c"
```
LionsAd: I have no idea how to get this to link, else ...

LionsAd:
@nikic Can you help with the windows failure? I have no idea what is wrong there ...

apc_lock.c
nikic: Adding the file in `config.m4` should be enough to link it. You have to rerun buildconf/phpize and configure for the changes to apply though.

apc_sma.c
```diff
@@ -59,6 +59,14 @@ struct sma_header_t {
 #define SMA_RO(sma, i)   ((char*)(sma->segs[i]).roaddr)
 #define SMA_LCK(sma, i)  ((SMA_HDR(sma, i))->sma_lock)
 
+#define SMA_CREATE_LOCK  APC_CREATE_MUTEX
+#define SMA_DESTROY_LOCK APC_DESTROY_MUTEX
+#define SMA_LOCK(sma, i) (APC_MUTEX_LOCK(&SMA_LCK(sma, i)))
+#define SMA_UNLOCK(sma, i) (APC_MUTEX_UNLOCK(&SMA_LCK(sma, i)))
+
+#define SMA_RLOCK(sma, i) (APC_MUTEX_RLOCK(&SMA_LCK(sma, i)))
+#define SMA_RUNLOCK(sma, i) (APC_MUTEX_RUNLOCK(&SMA_LCK(sma, i)))
```
nikic: I don't think keeping the read-lock distinction here is worthwhile. We should not expose read-lock ability in the mutex API (even if it is ultimately backed by an rwlock).

apc_sma.c
```diff
@@ -59,6 +59,14 @@ struct sma_header_t {
 #define SMA_RO(sma, i)   ((char*)(sma->segs[i]).roaddr)
 #define SMA_LCK(sma, i)  ((SMA_HDR(sma, i))->sma_lock)
 
+#define SMA_CREATE_LOCK  APC_CREATE_MUTEX
+#define SMA_DESTROY_LOCK APC_DESTROY_MUTEX
+#define SMA_LOCK(sma, i) (APC_MUTEX_LOCK(&SMA_LCK(sma, i)))
```
nikic: The outer parentheses here are redundant.

apc_mutex.c
```diff
@@ -0,0 +1,72 @@
...
+/* {{{ Cleanup attributes and statics */
+PHP_APCU_API void apc_mutex_cleanup() {
+	/* once per process please */
+	apc_mutex_ready = 0;
+	pthread_mutexattr_destroy(&apc_mutex_attr);
+}
+
+PHP_APCU_API zend_bool apc_mutex_create(apc_mutex_t *lock) {
+	pthread_mutex_init(lock, &apc_mutex_attr);
+        return 1;
```
nikic: nit: Indentation

apc_mutex.c
```diff
@@ -0,0 +1,72 @@
...
+}
+
+PHP_APCU_API zend_bool apc_mutex_unlock(apc_mutex_t *lock) {
+	pthread_mutex_unlock(lock);
+	HANDLE_UNBLOCK_INTERRUPTIONS();
+	return 1;
+}
+
+PHP_APCU_API void apc_mutex_destroy(apc_mutex_t *lock) {
+	// nothing to be done
```
nikic: Shouldn't this call `pthread_mutex_destroy`?

apc_mutex.c
```diff
@@ -0,0 +1,72 @@
...
+
+	pthread_mutexattr_settype(&apc_mutex_attr, PTHREAD_MUTEX_RECURSIVE);
+
+	return 1;
+}
+
+/* {{{ Cleanup attributes and statics */
+PHP_APCU_API void apc_mutex_cleanup() {
+	/* once per process please */
+	apc_mutex_ready = 0;
```
nikic: This variable doesn't seem to be doing anything? Shouldn't it be checked and set in apc_mutex_init?

apc_mutex.c
```diff
@@ -0,0 +1,72 @@
...
+	pthread_mutexattr_destroy(&apc_mutex_attr);
+}
+
+PHP_APCU_API zend_bool apc_mutex_create(apc_mutex_t *lock) {
+	pthread_mutex_init(lock, &apc_mutex_attr);
+        return 1;
+}
+
+PHP_APCU_API zend_bool apc_mutex_lock(apc_mutex_t *lock) {
+	int result;
```
nikic: This variable looks unused.

apc_mutex.c
```diff
@@ -0,0 +1,72 @@
...
+
+PHP_APCU_API zend_bool apc_mutex_init() {
+	if (pthread_mutexattr_init(&apc_mutex_attr) != SUCCESS) {
+		return 0;
+	}
+	if (pthread_mutexattr_setpshared(&apc_mutex_attr, PTHREAD_PROCESS_SHARED) != SUCCESS) {
+		return 0;
+	}
+
+	pthread_mutexattr_settype(&apc_mutex_attr, PTHREAD_MUTEX_RECURSIVE);
```
nikic: I don't think we need recursive mutexes, at least for the allocator.

apc_mutex.h
```diff
@@ -0,0 +1,65 @@
...
+#else
+
+#define apc_mutex_t apc_lock_t
+
+// Fallback to normal locks
+
+#define APC_MUTEX_INIT()          {}
+#define APC_MUTEX_CLEANUP()       {}
+
+#define APC_CREATE_MUTEX          CREATE_LOCK
```
nikic: All of these macros should pass through their argument. I think that will resolve the Windows build failure.

LionsAd:
Thanks for the review, I'll fix those issues today.

apc_lock.c
LionsAd: doh, my error was that I had the #ifdef without including apc.h / config.h, which means my implementation was never compiled.

apc_mutex.c
LionsAd: It does not in your lock cleanup, so I guess it's fine?

But yeah, seems like a bug in apc_lock.c. Will fix.

apc_sma.c
LionsAd: fixed

apc_mutex.c
LionsAd: fixed, still need to push that

LionsAd:
I added a commit to disable the destroying of the lock for now.

nikic:
Rebased and merged. Thanks a lot!

PR #340, opened 2018-11-18
Cleanup lock implementation

nikic:
Separate out each lock implementation, instead of interleaving all of them in deeply nested ifdef checks...

Also delete unused pgsql spinlock implementation.


LionsAd:
👍 I wanted to do exactly the same ...

PR #341, opened 2018-11-19
Clarify/fix cleanup operations

nikic:
Rename destroy/cleanup to detach to clarify that this is only about detaching local instances, not about cleaning up anything inside SHM. In particular remove calls to lock destruction.

Ref #338.


nikic:
/cc @LionsAd 

If nothing else, this is probably going to fix some issues on Windows, where the lock destroy implementation was notably not implemented as a noop.

LionsAd:
Looks great to me - the locks should be cleaned up automatically when the last forked process exits. The same for the sma region.

Issue #347, opened 2018-11-20
Broken 5.1.13 release apc_store()


thg2k:
Good morning,

with my utter disappointment release 5.1.13 broke my production. Last working release is 5.1.12 and I checked the diff and saw changes I'm not comfortable with in a **patch** release.

Are you guys scared of bumping major/minor versions? At least you give a signal that the upgrade should go through test staging before hitting production!

Anyway, this is the problem I'm facing: apc_store() randomly fails in this pre-check code:

```php
    /* try to store something to test if the cache is active */
    $_check = "fail";
    if (!apcu_store('th_cachesite_check', 42) ||
        (($_check = apcu_fetch('th_cachesite_check')) !== 42))
      throw new ThException("Site cache APCu supported but not enabled (check=\"$_check\")");
```

My setup is non-zts PHP 7.2.12 with apache prefork 2.4.37 and php as shared module.

nikic:
I'm sorry that the last release broke your production setup. The release was not intended to cause any significant behavior changes.

However, in this release the cache slam defense implementation, which has been broken for a while now, has been fixed. I strongly suspect that this is what causes the random apc_store() failures you are seeing. Can you please check whether setting the `apc.slam_defense=0` ini setting resolves your issue?

If you can confirm that this is the problem, I suggest that we switch the default of `apc.slam_defense` to off. While this setting has technically been enabled by default forever, in practice the slam defense was effectively off due to an implementation bug, and at this point it probably makes more sense to default it to off.

/cc @LionsAd

LionsAd:
Yes, I agree. We can properly re-enable cache defense in 5.2.0 as default with a proper announcement.

Technically apcu_store can fail at any time as there are many ways it can fail, so maybe adding a proper apcu_enabled function is the way to go so that OP can use an API function instead?

LionsAd:
Yes, I agree. We can properly re-enable cache defense in 5.2.0 as default with a proper announcement.

Technically apcu_store can fail at any time as there are many ways it can fail, so maybe adding a proper apcu_enabled function is the way to go so that OP can use an API function instead?

LionsAd:
Yes, I agree. We can properly re-enable cache defense in 5.2.0 as default with a proper announcement.

Technically apcu_store can fail at any time as there are many ways it can fail, so maybe adding a proper apcu_enabled function is the way to go so that OP can use an API function instead?

LionsAd:
Lol, seems like GH could use some slam detection.

[...]

PR #354, opened 2018-11-26
Try to reduce SMA fragmentation

nikic:
Instead of picking the first fitting block, check if there is
a smaller block that still fits. We only look at a limited, small
number of additional blocks. This is an approximation of best-fit
allocation.

This is intended to reduce fragmentation, in particular for the
case where a large block was recently freed (and is on top of the
free list) and a small allocation is performed.

### DIFF
--- a/apc_sma.c
+++ b/apc_sma.c
@@ -108,9 +108,48 @@ struct block_t {
 	#define RESET_CANARY(v)
 #endif
 
-/* {{{ MINBLOCKSIZE */
 #define MINBLOCKSIZE (ALIGNWORD(1) + ALIGNWORD(sizeof(block_t)))
-/* }}} */
+
+/* How many extra blocks to check for a better fit */
+#define BEST_FIT_LIMIT 3
+
+static inline block_t *find_block(sma_header_t *header, size_t realsize) {
+	void *shmaddr = header;
+	block_t *cur, *prv = BLOCKAT(ALIGNWORD(sizeof(sma_header_t)));
+	block_t *found = NULL;
+	uint32_t i;
+	CHECK_CANARY(prv);
+
+	while (prv->fnext) {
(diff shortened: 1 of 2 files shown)


nikic:
Possible fix for large allocation failures reported in #347.

Currently the allocator uses first-fit LIFO, which is known to result in very bad fragmentation. This patch is a very crude attempt to reduce fragmentation by using an approximate best-fit LIFO with short probe length.

The SMA allocator needs a lot more work, but I think this change is a safe and simple way to improve the situation.

/cc @LionsAd

LionsAd:
I am +1 to this change - Code looks great.

Only suggestion would be to not use a constant but a variable in the SMA so that we could make this configurable, e.g. Someone might need 5, while someone else only has many small allocations and needs the previous default of 1.

thelounge-zz:
BTW: does apcu catch concurrent writes?

apcu_store($key, $data, $ttl)

in case there is already a recor dwith the same key/value and it's TTL is not expired it should just return TRUE and skip any writes - i do that now in userland but only after the first apcu_store() failed to skip the retry because apcu_fetch() is not for free - well, it's also not in apcu itself but cheaper than php function calls and in case it helps to reduce fragmentation - maybe with an optin ini-setting?

$cache_rw = apcu_store($record_key, $content, $ttl);
if($cache_rw === FALSE)
{
 if(apcu_fetch($record_key) === $content)
 {
  return TRUE;
 }
 else
 {
  usleep(5);
  $cache_rw = apcu_store($record_key, $content, $ttl);
 }
}

nikic:
@thelounge-zz That's what apcu_add does. It stores the value if it doesn't exist (with non-expired TTL) yet.

LionsAd:
Yes, I'll open an issue for fetch-before-set with the same data. This should be hopefully easier once atomic updates are a thing.

PR #368, opened 2019-02-12
use mkstemp instead of mktemp

remicollet:
@nikic can you please have a look at mktemp usage which is strongly discouraged

I was able to fix 1 case, but unable to test it... can't find the bulid option to use this part of code... :(

Main issue is with apc_mmap.c, as only the name is needed and don't need to be a real path, can't find a "simple" fix for this one...

### DIFF
--- a/apc_lock.c
+++ b/apc_lock.c
@@ -281,8 +281,8 @@ PHP_APCU_API void apc_lock_cleanup() {
 
 PHP_APCU_API zend_bool apc_lock_create(apc_lock_t *lock) {
 	char lock_path[] = "/tmp/.apc.XXXXXX";
-	mktemp(lock_path);
-	*lock = open(lock_path, O_RDWR|O_CREAT, 0666);
+
+	*lock = mkstemp(lock_path);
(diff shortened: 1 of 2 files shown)


apc_lock.c
```diff
@@ -281,8 +281,12 @@ PHP_APCU_API void apc_lock_cleanup() {
 
 PHP_APCU_API zend_bool apc_lock_create(apc_lock_t *lock) {
 	char lock_path[] = "/tmp/.apc.XXXXXX";
+#if defined(HAVE_MKSTEMP)
```
nikic: As we already have unconditional usage in https://github.com/krakjoe/apcu/blob/master/apc_mmap.c#L111, we can use unconditionally here as well.

nikic:
I think you can just drop the `.shm` code in apc_mmap.c. If someone really cares, they can use a tmpfs to avoid syncing.

remicollet:
@nikic thanks, see new code

remicollet:
ty

PR #378, opened 2019-05-22
Removed deprecated uint and ulong types to conform to php master (8.0)

Antoine87:
The apcu extension does not compile on the current php master due to the removal of uint and ulong types.
See [the revelant commit](https://github.com/php/php-src/commit/bebcdcc74573b015c22238dbc9b2698c88b7a601).
And fixed another size_t/unsigned int conversion warning.

Untitled

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -577,7 +577,7 @@ static int apc_load_data(apc_cache_t* cache, const char *data_file)
 {
 	char *p;
 	char key[MAXPATHLEN] = {0,};
-	unsigned int key_len;
+	size_t key_len;
 	zval data;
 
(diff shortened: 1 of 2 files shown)


apc_cache.c
```diff
@@ -1095,7 +1095,7 @@ PHP_APCU_API zend_bool apc_cache_info(zval *info, apc_cache_t *cache, zend_bool
 					j++;
 				}
 				if (j != 0) {
-					add_index_long(&slots, (ulong)i, j);
+					add_index_long(&slots, (unsigned long)i, j);
```
nikic: `(zend_ulong)` would be better here.

nikic:
Thanks!

PR #385, opened 2019-10-08
Implement apcu_inc/dec with an atomic CAS loop

### LINKED ISSUE #343: Use atomic operations for apcu_inc/dec
This was already implemented in #334 for apcu_cas. The goal is to allow these operations to be performed under a read-lock instead of a write-lock.

A complication for apcu_inc/dec is that we'd have to switch to wraparound behavior (rather than saturating to a float). At least if we want to use straight line atomic add/sub rather than a cas loop with explicit overflow check.

nikic:
This is an alternative to #384, which uses an atomic CAS loop instead of an atomic add to preserve the current saturation to double. I'm not sure doing this makes sense.

Fixes #343.

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -901,11 +901,11 @@ retry_update:
 
 /* {{{ apc_cache_atomic_update_long */
 PHP_APCU_API zend_bool apc_cache_atomic_update_long(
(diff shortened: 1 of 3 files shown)


nikic:
Closing this in favor of #384.

PR #389, opened 2019-10-30
Add new method:apcu_exists_any

fuzhufang:
apcu_exists_any: In the array parameters passed in, look up in order, as soon as you find one, exit immediately, return true.  Application scenario: You only need to know if one of all the keys exists, you don't need to know if all the keys exist, and you don't need to know who the keys are.


nikic:
Do you have an example where this would be used? I'm wondering if it wouldn't be better to at least return which of the keys is the one that exists.

fuzhufang:
> Do you have an example where this would be used? I'm wondering if it wouldn't be better to at least return which of the keys is the one that exists.

For example: I have an IP whitelist and a device whitelist and even more conditions. Every time the user accesses the service, we need to check whether it is an IP whitelist or whether it is a device whitelist or other conditions, but I Just need to know if one is established. I think using apcu_exists can also satisfy this requirement, but it would be more efficient if you use apcu_exists_any. Or you have a better solution

I agree that it is better to return which key is the key that exists, thank you for your comment.

stof:
if the implementation only loops synchronously over the keys to check if the key exists, is there any benefit about doing that in the extension vs doing it in userland ?

Issue #391, opened 2020-03-10
"apc.use_request_time" - Better understand or reconsider default


Krinkle:
Hi,

For Wikipedia's deployment, and for the MediaWiki software in general, we're currently considering to turn this off everywhere ([patch](https://gerrit.wikimedia.org/r/#/c/mediawiki/core/+/578400/)).

-------

> If a value is saved with a TTL of X, then it should take ~X seconds for it to expire. A key may need to use low, often randomized, TTLs, and this behavior breaks such use cases.
> 
> The default `apc.use_request_time` value (1) assigns writes a timestamp equal to the web request timestamp. If a request has been going for Y seconds, then a TTL of X becomes (X-Y) as far as new requests can see, due to the "back-dating". If Y > X, then the value was effectively purged for all new requests. This can easily be tested via `sleep()`.

-------

I'd like to better understand why this setting is enabled by default. If there are cases we may've missed, I'd like to help document these better for the benefit of others (and to perhaps keep the default the afterall).

If not, perhaps it would make sense for the next major version to disable this as well.

Thanks!

/cc @AaronSchulz

nikic:
I believe this is just a performance consideration, to avoid calling time() on every apc API call. I have no idea if this is still relevant nowadays or not.

AaronSchulz:
From some quick profiling, I didn't notice a difference.
```
function testAPC( $use_req_time ) {
	ini_set( 'apc.use_request_time', (string)$use_req_time );

	$t = microtime( true );
	for ( $i = 0; $i < 1000000; ++$i ) {
		if ( !apcu_store( 'x', 'a', 10 ) ) {
			throw new Exception( "Failed" );
		}
	}
	var_dump( 'store', microtime( true ) - $t );

	$t = microtime( true );
	for ( $i = 0; $i < 1000000; ++$i ) {
		if ( apcu_fetch( 'x' ) === false ) {
			throw new Exception( "Failed" );
		}
	}
	var_dump( 'get', microtime( true ) - $t );
}
```

```
>>> testAPC(0);
string(5) "store"
float(0.75342202186584)
string(3) "get"
float(0.38800597190857)
=> null
>>> testAPC(1);
string(5) "store"
float(0.7548041343689)
string(3) "get"
float(0.38969397544861)
=> null
```

PR #392, opened 2020-03-11
Set apc.use_request_time=0 by default

Krinkle:
The default used to be 1, which meant that the effective TTL upon storing and fetching keys was backdated to the beginning of the web request.

This means values are not stored as long as users might expect, and might also appear valid longer than expected.

For example, during POST request that 2s and stores a value in APCu with a TTL of 500ms, the value effectively can't be re-used by other requests as 0.5 - 2 <= 0.

Closes https://github.com/krakjoe/apcu/issues/391.


nikic:
Seems like the better default. I don't see a perf difference either, and people who do can enable explicitly.

PR #395, opened 2020-05-03
Use stable version of PHP 7.4 and fix config warnings

carusogabriel:
These config warnings are available in the [`View Config`](https://travis-ci.org/github/krakjoe/apcu/builds/682281522/config) tab on any of the Travis' builds.


carusogabriel:
cc @nikic

andypost (approved):

PR #396, opened 2020-06-08
Improve and update AppVeyor CI integration

cmb69:
This PR is mainly about updating the PHP versions for the builds (was 7.0-7.2, now 7.2-7.4), but also improves maintainability of the AppVeyor script, and fixes the test failures on PHP 7.4.


nikic:
As the ext still supports 7.0, might it make more sense to test 7.0-7.4, but drop some of the configs? Just test x64 nts for old versions, for example.

cmb69:
I've added jobs for 7.0 and 7.1 (x64 nts).

PR #431, opened 2022-01-31
Raise error on invalid serializer.

zeriyoshi:
APCu works with the default serializer, even if the value set in apc.serializer is invalid.

But, this is unfriendly if you are using a third party serializer such as igbinary. For example, if igbinary is installed before APCu is installed, igbinary will not support APCu. The current behavior of working with the default serializer without warning may produce unintended behavior.

This fix changes it to give an E_WARNING equivalent warning if an unavailable serializer is specified.

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -1218,6 +1218,9 @@ PHP_APCU_API zend_bool apc_cache_defense(apc_cache_t *cache, zend_string *key, t
 PHP_APCU_API void apc_cache_serializer(apc_cache_t* cache, const char* name) {
 	if (cache && !cache->serializer) {
 		cache->serializer = apc_find_serializer(name);
+		if (strcmp(name, "default") != 0 && !cache->serializer) {
+			php_error_docref(NULL, E_WARNING, "apc_cache_serializer: serializer \"%s\" is not supported.", name);
+		}
 	}
 } /* }}} */
 
--- a/tests/serializer_valid.phpt
+++ b/tests/serializer_valid.phpt
@@ -1,0 +1,11 @@
+--TEST--
+Serializer: valid pattern.
+--SKIPIF--
+<?php require_once(dirname(__FILE__) . '/skipif.inc'); ?>
+--INI--
+apc.enabled=1
+apc.enable_cli=1
+apc.serializer=php
+--FILE--
+<?php
+--EXPECT--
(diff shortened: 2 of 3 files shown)


apc_cache.c
```diff
@@ -1218,6 +1218,9 @@ PHP_APCU_API zend_bool apc_cache_defense(apc_cache_t *cache, zend_string *key, t
 PHP_APCU_API void apc_cache_serializer(apc_cache_t* cache, const char* name) {
 	if (cache && !cache->serializer) {
 		cache->serializer = apc_find_serializer(name);
+		if (strcmp(name, "default") != 0 && !cache->serializer) {
+			php_error_docref(NULL, E_WARNING, "apc_cache_serializer: serializer \"%s\" is not supported.", name);
```
nikic, line 1222:
```suggestion
			php_error_docref(NULL, E_WARNING, "apc_cache_serializer: serializer \"%s\" is not supported", name);
```

tests/serializer_valid.phpt
```diff
@@ -0,0 +1,11 @@
+--TEST--
```
TysonAndre, line 1: This seems to be redundant and the various other tests already assert this notice isn't emitted for valid serializers

apc_cache.c
```diff
@@ -1218,6 +1218,9 @@ PHP_APCU_API zend_bool apc_cache_defense(apc_cache_t *cache, zend_string *key, t
 PHP_APCU_API void apc_cache_serializer(apc_cache_t* cache, const char* name) {
 	if (cache && !cache->serializer) {
 		cache->serializer = apc_find_serializer(name);
+		if (strcmp(name, "default") != 0 && !cache->serializer) {
+			php_error_docref(NULL, E_WARNING, "apc_cache_serializer: serializer \"%s\" is not supported", name);
```
TysonAndre, line 1222: Not a maintainer but I'm somewhat familiar with the codebase

`not supported or not enabled`, maybe. There's also the question of whether it should fall back to the "php" serializer instead, given that the actual default ini value is "php" (see linked ticket)

zeriyoshi: Indeed. The current behavior is complicated and cannot be understood without following the implementation. Also, the situation where the fallback behavior that occurs when it is not present and the flow when php is explicitly specified are separated is probably not desirable.

I have added a comment to your Issue.

PR #433, opened 2022-02-07
PHP 8.1 compatibility

zebroid:
1) without rounding PHP 8.1 generates A LOT of errors like "PHP Deprecated:  Implicit conversion from float 115.87685818391442 to int loses precision" in GDI functions (imagefilledarc, imagearc, imageline and so on)

2)  some settings (such as "apc.mmap_file_mask" or "apc.preload_path") may not have any values at all, so PHP 8.1 generate error "PHP Deprecated:  str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated" in enumeration  "foreach (ini_get_all('apcu') as $k => $v)"

### DIFF
--- a/apc.php
+++ b/apc.php
@@ -228,48 +228,53 @@ if (isset($MYREQUEST['IMG']))
 
 	function fill_arc($im, $centerX, $centerY, $diameter, $start, $end, $color1,$color2,$text='',$placeindex=0) {
 		$r=$diameter/2;
-		$w=deg2rad((360+$start+($end-$start)/2)%360);
-
+		$w=deg2rad(round(360+$start+($end-$start)/2)%360);
 
 		if (function_exists("imagefilledarc")) {
 			// exists only if GD 2.0.1 is available
-			imagefilledarc($im, $centerX+1, $centerY+1, $diameter, $diameter, $start, $end, $color1, IMG_ARC_PIE);
-			imagefilledarc($im, $centerX, $centerY, $diameter, $diameter, $start, $end, $color2, IMG_ARC_PIE);
-			imagefilledarc($im, $centerX, $centerY, $diameter, $diameter, $start, $end, $color1, IMG_ARC_NOFILL|IMG_ARC_EDGED);
+			imagefilledarc($im, $centerX+1, $centerY+1, $diameter, $diameter, round($start), round($end), $color1, IMG_ARC_PIE);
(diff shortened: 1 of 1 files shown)


apc.php
```diff
@@ -807,7 +812,9 @@ function block_sort($array1, $array2)
 
 	$j = 0;
 	foreach (ini_get_all('apcu') as $k => $v) {
-		echo "<tr class=tr-$j><td class=td-0>",$k,"</td><td>",str_replace(',',',<br />',$v['local_value']),"</td></tr>\n";
+        //some settings (such as "apc.mmap_file_mask" or "apc.preload_path") may not have any values at all
+        //so PHP 8.1 generate error "PHP Deprecated:  str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated"
+		echo "<tr class=tr-$j><td class=td-0>",$k,"</td><td>",(isset($v['local_value']) ? str_replace(',',',<br />',$v['local_value']) : ''),"</td></tr>\n";
```
nikic, line 817: Maybe using `$v['local_value'] ?? ''` would be more elegant here? I don't think the comment is necessary.

andypost:
still useful

zebroid:
> still useful

so, maybe you can answer what to do?

"I couldn't find what is the minimum PHP version that apc.php supports. Of course "$v['local_value'] ?? ''" is more elegant (for PHP >= 7.0).

Also I am new to github. Should I submit a new version of the patch, or you can change these lines by yourself?"

andypost:
I bet https://www.php.net/supported-versions.php is primary source of truth for supported versions

So it's totally fine for 7.4

andypost:
Btw https://github.com/krakjoe/apcu/commit/2bde1c8d794c787a610b8226c399626967611721 declared 7.2

So better to wait for @nikic POV

nikic:
apcu currently supports PHP >= 7.0.

apc.php
andypost: so it can't be used as package could be used on 7.0+

andypost: @nikic pleae merge it as is

nikic: Null coalesce was added in PHP 7.0, so I don't see the problem.

andypost: Fixed in amended commit https://github.com/krakjoe/apcu/pull/444

andypost: if the #444 will be merged this one should be closed

nikic:
Closing in favor of #444.

PR #436, opened 2022-03-23
Enable compiler warnings & fix them

MaxKellermann:


MaxKellermann:
CI failures unrelated to this PR, again.

config.m4
```diff
@@ -231,6 +231,10 @@ if test "$PHP_APCU" != "no"; then
   		[AC_DEFINE([HAVE_VALGRIND_MEMCHECK_H],1, [enable valgrind memchecks])])
   ])
 
+  for i in -Wall -Wextra -Wno-unused-parameter; do
+    AX_CHECK_COMPILE_FLAG([$i], [APCU_CFLAGS="$APCU_CFLAGS $i"])
```
nikic, line 235: I'm not entirely sure whether AX_CHECK_COMPILE_FLAG is available on all PHP versions apcu supports. Do we have it in 7.0? Annoying that there is no working CI.

nikic: Looks like it is present: https://github.com/php/php-src/blob/PHP-7.0/build/ax_check_compile_flag.m4

MaxKellermann: I wish there was `AX_APPEND_COMPILE_FLAGS` ... I had to roll my own here. Would it be appropriate to add autoconf-archive files to the apcu repository?

nikic: Don't think we need that to save the one line. Not sure we really need the check at all, as these are pretty standard flags anyway.

PR #441, opened 2022-07-18
Fix AppVeyor CI build for PHP 7.3

cmb69:
PHP 7.3 is EOL, so the builds have been moved to the archives; we need
to cater to that.

---

An alternative might be to drop AppVeyor CI in favor of GH actions altogether.


nikic:
> An alternative might be to drop AppVeyor CI in favor of GH actions altogether.

I think the main thing AppVeyor adds is that it produces artifacts, which is useful to have before the mass rebuild... Maybe this is possible with Github Actions as well?

PR #449, opened 2022-09-11
Remove Travis CI config

pmurzakov:


nikic:
We no longer use Travis -- so feel free to delete the config file instead :)

pmurzakov:
Ah, sorry, I didn't notice it.
I've just removed the config.

Shall we update Appveyor to include PHP 8.1 then?

nikic:
> Shall we update Appveyor to include PHP 8.1 then?

I think we can drop Appveyor as well, as we have Windows testing on GH actions now. The only thing it adds is x86 builds. We could add those on the GH actions side as well, would just make the matrix larger.

PR #464, opened 2022-11-05
Add editorconfig file for APCu

TysonAndre:
https://editorconfig.org/ has integrations with many editors/IDEs to apply project-specific settings (e.g. to consistently use tabs)

https://www.php-fig.org/psr/psr-12/ recommends using spaces for php code. (though many of these tests use tabs, I don't have a strong preference)

Replace the mix of tabs and spaces in config.m4 with spaces.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+; 4 space indentation (no size specified)
+[*]
+charset = utf-8
+indent_style = tab
+indent_size = 4
+[*.{php,c,h,m4,w32,md}]
+trim_trailing_whitespace = true
+[*.{phpt,php}]
+indent_style = space
+[*.xml]
+indent_size = 1
+indent_style = space
+[*.m4]
+indent_size = 2
+indent_style = space
+[*.w32]
+charset = utf-8
+indent_style = tab
+indent_size = 2
--- a/apc_mutex.h
+++ b/apc_mutex.h
@@ -50,8 +50,8 @@ typedef apc_lock_t apc_mutex_t;
 
 // Fallback to normal locks
 
-#define APC_MUTEX_INIT()          
-#define APC_MUTEX_CLEANUP()       
+#define APC_MUTEX_INIT()
+#define APC_MUTEX_CLEANUP()
 
 #define APC_CREATE_MUTEX(lock)    CREATE_LOCK(lock)
 #define APC_DESTROY_MUTEX(lock)   DESTROY_LOCK(lock)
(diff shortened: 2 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+indent_style = space
+[*.xml]
+indent_size = 1
+indent_style = space
+[*.m4]
+indent_size = 2
+indent_style = space
+[*.w32]
+charset = utf-8
+end_of_line = crlf
```
nikic, line 19: Is this really needed and not handled by git checkout?

TysonAndre: > Is this really needed and not handled by git checkout?

I'm not sure what you mean.

`*` was added to use line feed for all other types of files (e.g. brand new files added by contributors), this is added to override that for the line endings currently used by config.w32. (`\r\n`)

Some editors/IDEs do a bad job with pasting in snippets copied from elsewhere and can end up with mixes of newline types

nikic: Well, why is this commit changing that file? That shouldn't happen.

nikic: Most likely end_of_line just shouldn't be specified in editorconfig at all, and left up to git.

TysonAndre: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings are you referring to this feature of git? https://github.com/krakjoe/apcu doesn't have a `.gitattributes` file.

config.w32
```diff
@@ -1,32 +1,32 @@
...
+
+	/* XXX srwlock kernel as most BC compatible for now */
+	AC_DEFINE('APC_SRWLOCK_KERNEL', 1);
+	AC_DEFINE('HAVE_APCU', 1);
+	ADD_FLAG('CFLAGS_APCU', '/D WIN32_ONLY_COMPILER=1 /DAPC_SRWLOCK_KERNEL=1');
+	PHP_INSTALL_HEADERS("ext/apcu", "php_apc.h apc.h apc_api.h apc_cache.h apc_globals.h apc_iterator.h apc_lock.h apc_mutex.h apc_sma.h apc_serializer.h apc_stack.h apc_windows_srwlock_kernel.h apc_arginfo.h php_apc_legacy_arginfo.h");
+
+	EXTENSION('apcu', apc_sources, PHP_APCU_SHARED, "/DZEND_ENABLE_STATIC_TSRMLS_CACHE=1");
+}
+
```
nikic, line 32: The changes to this file still need to be dropped.

.editorconfig
nikic: I'm referring to the general core.autocrlf setting, which people can use to convert LF to CRLF in Windows checkouts, if necessary.

config.w32
TysonAndre: Done

.editorconfig
TysonAndre: Removed the end_of_line override

Issue #323, opened 2018-10-08
Optimization ideas for persistence logic


nikic:
A number of minor improvements that could be made to the new persistence mechanism implemented in #305:

 * [ ] Memoize strings used in array keys. We memoize other strings.
 * [ ] Compact arrays if they have many holes.
 * [ ] Tag pointers that have been memoized. This should be pretty rare and we can avoid HT lookups for pointers that aren't tagged.
 * [ ] String structures can be copied in their entirety, we don't have to initialize everything separately.

TysonAndre:
>  Memoize strings used in array keys. We memoize other strings.

If anyone else implements this: I believe this means within the representation of a single apcu entry (with apc.serializer=default , used for arrays not containing objects). String reuse is safely done in the igbinary unserializer when unserializing arrays (and objects) https://github.com/igbinary/igbinary - `igbinary.compact_strings` ini section

> String structures can be copied in their entirety, we don't have to initialize everything separately.

If anyone else implements this: I assume that means a memcpy on the entire zend_string from shared memory into the emalloced temporary string - the original string in shared memory is propery initialized, so the memcpy would be as well.

-------

These optimization ideas still look to be correct in the latest php versions for temporary strings.

Having an option for an immutable data store that can't be cleared or have entries replaced would allow for much more aggressive optimizations - See https://github.com/krakjoe/apcu/issues/175 

- Though opcache_add_value(string $key, mixed $value): bool and opcache_load_value(string $key): mixed may be a better place for me to propose that from what I was looking into in 175, since opcache is already storing persistent strings and php values (including file cache) (ini settings could limit the memory this could take up to a fraction of total opcache memory, leaving space for php scripts)

TysonAndre:
Other optimization ideas:

- [ ] zend_hash will hash the index integers to `% power_of_2_size`. Pointers such as zvals are 128 bits long (16 bytes) and aligned to at least 8 bytes, so they all hash to the same 1 in 64 hash buckets in practice (may be more), causing a likely noticeable slowdown due to the many hash collisions for large values.  Shift pointers by ZEND_MM_ALIGNMENT_LOG2?

 See https://github.com/php/php-src/pull/7690 

```c
static inline void *apc_persist_get_already_allocated(apc_persist_context_t *ctxt, void *ptr) {
	if (ctxt->memoization_needed) {
		return zend_hash_index_find_ptr(&ctxt->already_allocated, (uintptr_t) ptr);
	}
	return NULL;
}
```

[...]

PR #466, opened 2022-11-05
Reduce shared memory usage storing the empty array

TysonAndre:
Reduce shared memory usage in php 7.3+ for `apc.serializer=default`. Store the pointer to the immutable zend_empty_array rather than a brand new array. (APCu is already storing a pointer to static variables elsewhere, e.g. uninitialized_bucket)

### DIFF
--- a/apc_persist.c
+++ b/apc_persist.c
@@ -127,10 +127,14 @@ static zend_bool apc_persist_calc_memoize(apc_persist_context_t *ctxt, void *ptr
 static zend_bool apc_persist_calc_ht(apc_persist_context_t *ctxt, const HashTable *ht) {
 	uint32_t idx;
 
-	ADD_SIZE(sizeof(HashTable));
-	if (ht->nNumUsed == 0) {
+	/* In php 7.3+, this points to the immutable zend_empty_array outside of shared memory. */
+	if (ht->nNumOfElements == 0) {
+#if PHP_VERSION_ID < 70300
+		ADD_SIZE(sizeof(HashTable));
+#endif
 		return 1;
 	}
+	ADD_SIZE(sizeof(HashTable));
 
 	/* TODO Too sparse hashtables could be compacted here */
 #if PHP_VERSION_ID >= 80200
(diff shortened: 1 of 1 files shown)


nikic:
I don't think this is correct in this form. Not every empty array is a zend_empty_array.

Another thing to keep in mind is that empty arrays have have different next free elements, though I'd have to check whether apcu preserves that, possibly it's already ignored.

TysonAndre:
I fixed the error - the earlier commit made this use 0 memory, but didn't set the zval to zend_empty_array when persisting.


> I don't think this is correct in this form. Not every empty array is a zend_empty_array.

I know, but the question is whether any empty array (in userland arrays) can be replaced with the zend_empty_array without observable effects. On unpersist, this is already using the EMPTY_ARRAY from #323 

```c
		case IS_ARRAY:
#if PHP_VERSION_ID >= 70300
			if (Z_ARR_P(zv)->nNumOfElements == 0) {
				ZVAL_EMPTY_ARRAY(zv); /* #323 */
				return;
			}
#endif
			Z_ARR_P(zv) = apc_unpersist_ht(ctxt, Z_ARR_P(zv));
			return;
```

> Another thing to keep in mind is that empty arrays have have different next free elements, though I'd have to check whether apcu preserves that, possibly it's already ignored.

apc_persist_copy_ht ignores that. serialize() functions also ignore that by design.

```c
	if (ht->nNumUsed == 0) {
#if PHP_VERSION_ID >= 70400
		ht->u.flags = HASH_FLAG_UNINITIALIZED;
#else
		ht->u.flags &= ~(HASH_FLAG_INITIALIZED|HASH_FLAG_PACKED);
#endif
		ht->nNextFreeElement = 0;
		ht->nTableMask = HT_MIN_MASK;
		HT_SET_DATA_ADDR(ht, &uninitialized_bucket);
		return ht;
	}
```

apc_persist.c
```diff
@@ -127,10 +127,16 @@ static zend_bool apc_persist_calc_memoize(apc_persist_context_t *ctxt, void *ptr
 static zend_bool apc_persist_calc_ht(apc_persist_context_t *ctxt, const HashTable *ht) {
 	uint32_t idx;
 
+#if PHP_VERSION_ID < 70300
 	ADD_SIZE(sizeof(HashTable));
-	if (ht->nNumUsed == 0) {
+#endif
+	if (ht->nNumOfElements == 0) {
```
nikic, line 133: This is written in a confusing way. I'd move the condition ADD_SIZE in here...

apc_persist.c
```diff
@@ -127,10 +127,16 @@ static zend_bool apc_persist_calc_memoize(apc_persist_context_t *ctxt, void *ptr
...
 		return 1;
 	}
+#if PHP_VERSION_ID >= 70300
+	/* In php 7.3+, this points to the immutable zend_empty_array outside of shared memory. */
+	ADD_SIZE(sizeof(HashTable));
```
nikic, line 138: ...and make this one unconditional.

PR #467, opened 2022-11-05
Update TECHNOTES

### LINKED ISSUE #452: TECHNOTES.md is out of date
1. This refers to CVS on a missing server, not github
2. Struct definitions are mostly the same as when it was created some property types changed, e.g. block_t has extra fields now for the previous block
3. Some functions are out of date, e.g. make_slot/apc_cache_insert don't exist. TSRMLS_CC/TSRMLS_DC is no longer needed in php 7
4. The comment about powers of two and POWER_OF_TWO_BLOCKSIZE seems incorrect/out of date, in `static APC_HOTSPOT size_t sma_allocate(sma_header_t *header, size_t size, size_t fragment, size_t *allocated)` and `find_block` it's just finding the first block 
5. I'm not familiar enough with the architecture to be sure if the description is correct

TysonAndre:
Closes #452

### DIFF
--- a/TECHNOTES.txt
+++ b/TECHNOTES.txt
@@ -1,28 +1,25 @@
-APC Quick-Start Braindump
+APCu Quick-Start Braindump
 
-This is a rapidly written braindump of how APC currently works in the
-form of a quick-start guide to start hacking on APC.
+This is a rapidly written braindump of how APCu currently works in the
+form of a quick-start guide to start hacking on APCu.
 
 1. Install and use APC a bit so you know what it does from the end-user's
    perspective.  
-   user-space functions are all explained here: 
+   user-space functions are all explained here: https://www.php.net/apcu
 
-2. Grab the current APC code from CVS:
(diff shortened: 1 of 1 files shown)


TECHNOTES.txt
```diff
@@ -38,37 +35,37 @@ form of a quick-start guide to start hacking on APC.
...
+        2) pooling
+        3) user land cache
+
 5.1) APCu SMA
    
    It is a pretty standard memory allocator, now supporting third party extensions.
 
-   apc_sma_malloc, apc_sma_realloc, apc_sma_strdup and apc_sma_free 
-   behave to the caller just like malloc, realloc, strdup and free, they are 
+   apc_sma_malloc, apc_sma_malloc, apc_sma_free, etc
```
nikic, line 46: apc_sma_malloc listed twice?

TysonAndre: Fixed

PR #469, opened 2022-11-12
Fix error linking with pthread

TysonAndre:
Noticed in php 8.3-dev configured with --disable-all

The "-pthread" flag wasn't actually getting used in the makefile.


config.m4
```diff
@@ -122,7 +122,7 @@ if test "$PHP_APCU" != "no"; then
           }
         ]])],[ dnl -Success-
           APCU_CFLAGS="-D_GNU_SOURCE -DZEND_ENABLE_STATIC_TSRMLS_CACHE=1"
-          PHP_ADD_LIBRARY(pthread)
+          PHP_ADD_LIBRARY(pthread,, APCU_SHARED_LIBADD)
           PHP_LDFLAGS="$PHP_LDFLAGS -lpthread"
```
nikic, line 126: Hm, do we need the PHP_LDFLAGS adjustment? As I understand it, if apcu is not build shared PHP_ADD_LIBRARY should already be adding it to LIBS.

TysonAndre: I'm assuming not, and can't think of any reason to have this, but not 100% sure and haven't checked back to 7.0. It gets added to the Makefile but nothing in the Makefile uses PHP_LDFLAGS.

Even when compiling apcu statically this still wouldn't affect other modules

nikic: In that case I think we should drop these lines and see if anyone complains...

PR #475, opened 2022-11-13
Also reject negative format values in APCuIterator

TysonAndre:
This would affect applications that accidentally or deliberately pass in negative values for $format (e.g. PHP_INT_MIN on 64-bit builds)

```c
	if (format > APC_ITER_ALL) { /* 0xffffffff */
		apc_error("APCUIterator format is invalid");
		return;
	}
```


nikic:
Doesn't looks that this error condition is tested currently, add one?

nikic:
Looks like apc_099.phpt is failing.

Issue #365, opened 2019-01-19
apc.php request too much memory exceeding php memory_limit


PutinVladimir:
When cached 1500000 entries apc.php request more then 1500mb at execution.

Issue is apcu_cache_info.

FastCGI sent in stderr: "PHP message: PHP Fatal error:  Allowed memory size of 402653184 bytes exhausted (tried to allocate 20480 bytes) in apcu.php on line 190"

if set php memory_limit 1600mb or  if set $limited = true it works.

Djibs:
I have the same problem

chrishecker:
For me the solution to this was putting this at the top of apc.php (after the <?php line obvs):
ini_set('memory_limit','256M');

The amount should be bigger than the size that's failing, so probably 512M for you.  I'm guessing 2x the apc.shm_size would work, so I was going to write a thing that got that and set the limit appropriately, but haven't done that yet.

Chris

TysonAndre:
It seems like https://www.php.net/apcu_cache_info should instead be called with `$limited = true` in apc.php when the request isn't for the OB_USER_CACHE table.

By default, apcu_cache_info() returns an array corresponding to each entry

```
// operation constants
define('OB_HOST_STATS',1);
define('OB_USER_CACHE',2);
define('OB_VERSION_CHECK',3);
// ...
// check validity of input variables
$vardom=array(
	'OB'	=> '/^\d+$/',			// operational mode switch
// ...
if (empty($MYREQUEST['OB']))	$MYREQUEST['OB']=OB_HOST_STATS;

```

madmajestro:
This has been resolved via #477 a few years ago. I'm closing this.

PR #477, opened 2022-11-14
Reduce memory usage of apc.php. Fix division by zero error.

TysonAndre:
On tabs other than "User Cache Entries", don't fetch metadata about cache entries.

When running php apc.php in CLI to preview this (adding a cached entry),
I noticed a DivisionByZeroError this fixes.

For https://github.com/krakjoe/apcu/issues/365

## Testing

https://www.php.net/manual/en/function.apcu-cache-info.php#refsect1-function.apcu-cache-info-changelog 

I previewed this in cli for convenience by using the environment instead of an http request

`for IMG in {1..3}; do export IMG; php apc.php > $IMG.png; done` - the image generation doesn't use apcu_cache_info.

```diff
--- a/apc.php
+++ b/apc.php
@@ -95,6 +95,10 @@ $scope_list=array(
        'A' => 'cache_list',
        'D' => 'deleted_list'
 );
+$_REQUEST = $_ENV;
+for ($i = 0; $i < 21; $i++) {
+       apcu_store("k$i", str_repeat('x', 2**$i));
+}
 
```

### DIFF
--- a/apc.php
+++ b/apc.php
@@ -183,7 +183,8 @@ if(!function_exists('apcu_cache_info')) {
   exit;
 }
 
-$cache = apcu_cache_info();
+$limited = $MYREQUEST['OB'] != OB_USER_CACHE;
+$cache = apcu_cache_info($limited);
 
 $mem=apcu_sma_info();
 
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -7,6 +7,8 @@ indent_size = 4
 trim_trailing_whitespace = true
 [*.{phpt,php}]
 indent_style = space
+[apc.php]
+indent_style = tab
```
nikic, line 11: Would you mind (in a separate PR) reformatting the file to spaces? We should not use tabs in PHP files.

apc.php
```diff
@@ -183,7 +183,8 @@ function defaults($d,$v) {
   exit;
 }
 
-$cache = apcu_cache_info();
+$limited = $MYREQUEST['OB'] != OB_USER_CACHE;
+$cache = version_compare(phpversion('apcu'), '3.0.11', '>=') ? apcu_cache_info($limited) : apcu_cache_info();
```
nikic, line 187: 3.0.11 is extremely old, I don't think we need this check.

.editorconfig
TysonAndre: I split it into a separate commit so I don't get merge conflicts when I rebase, #478 has just the reformatting

PR #492, opened 2023-07-07
feat: improve search pattern validation

hrvoj3e:
One should be able to use dash (`-`), wildcard (`*`) and backslash (`\`) in search.

E.g. if a key `myproject.MyNsA\MyNsB\MyClassC-dataMap` exists in cache  I would search `myproject\.MyNsA\\MyNsB.*-dataMap` and get only that specific pattern.

### DIFF
--- a/apc.php
+++ b/apc.php
@@ -87,7 +87,7 @@ $vardom=array(
     'SORT1' => '/^[AHSMCDTZ]$/',    // first sort key
     'SORT2' => '/^[DA]$/',          // second sort key
     'AGGR'  => '/^\d+$/',           // aggregation by dir level
-    'SEARCH'    => '~^[a-zA-Z0-9/_.-]*$~'           // aggregation by dir level
+    'SEARCH' => '~^[a-zA-Z0-9/_.\-*\\\]*$~' // search term regex
 );
 
 // cache scope


apc.php
```diff
@@ -87,7 +87,7 @@ function defaults($d,$v) {
     'SORT1' => '/^[AHSMCDTZ]$/',    // first sort key
     'SORT2' => '/^[DA]$/',          // second sort key
     'AGGR'  => '/^\d+$/',           // aggregation by dir level
-    'SEARCH'    => '~^[a-zA-Z0-9/_.-]*$~'           // aggregation by dir level
+    'SEARCH' => '~^[a-zA-Z0-9/_.\-*\\\]*$~' // search term regex
```
nikic, line 90: Any reason not to allow arbitrary inputs?

hrvoj3e: Arbitrary will pass this validation and preg_match that is used for searching. This has worked before and will now.

The problem are specific regex inputs that will not pass validation. I provided a complex example from our app that we use in cache and have not been able to search it.

hrvoj3e: I just want to say that this PR does not break current behavior in any way - it just improves searching via regex.

PR #507, opened 2024-03-29
Feature/apcu_set_ttl Add function to set the TTL of an existing entry atomically

ignas2526:
APCu lacks a nice way to update the TTL of an entry. Currently, you'd have to implement your own critical section, fetch the value of an entry, and then store it back. I think there are many use cases with websites where one would want to create a cache entry with a short life and then extend its lifespan on a subsequent request from the browser.

I think this will be especially useful in combination with apcu_inc, apcu_dec, and apcu_cas. Because calling apcu_store (the only way to extend TTL ATM) after them defeats their purpose.

This PR adds `apcu_set_ttl($key, int $ttl = 0): bool`. If entry with the $key has been successfully updated, function returns true, else, false.

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -273,7 +273,7 @@ PHP_APCU_API int APC_UNSERIALIZER_NAME(php) (APC_UNSERIALIZER_ARGS)
 	result = php_var_unserialize(value, &tmp, buf + buf_len, &var_hash);
 	PHP_VAR_UNSERIALIZE_DESTROY(var_hash);
 	BG(serialize_lock)--;
-	
+
 	if (!result) {
 		php_error_docref(NULL, E_NOTICE, "Error at offset %td of %zd bytes", tmp - buf, buf_len);
 		ZVAL_NULL(value);
@@ -537,6 +537,33 @@ PHP_APCU_API zend_bool apc_cache_store(
 	}
 
 	return ret;
+} /* }}} */
+
+/* {{{ apc_cache_set_ttl */
+PHP_APCU_API zend_bool apc_cache_update_ttl(
+		apc_cache_t* cache, zend_string *key, const int32_t ttl) {
+	apc_cache_entry_t *entry;
+	zend_bool retval = 0;
+	time_t t = apc_time();
+
+	if (!cache) {
+		return 0;
+	}
+
+	if (!apc_cache_rlock(cache)) {
+		return 0;
+	}
+
+	entry = apc_cache_rlocked_find_nostat(cache, key, t);
+	if (!entry) {
+		apc_cache_runlock(cache);
+		return 0
+	}
+
+	entry->ttl = ttl;
+
+	apc_cache_runlock(cache);
+	return 1;
 } /* }}} */
 
 #ifndef ZTS
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -560,6 +560,7 @@ PHP_APCU_API zend_bool apc_cache_update_ttl(
 		return 0;
 	}
 
+	entry->ctime = t;
 	entry->ttl = ttl;
 
 	apc_cache_runlock(cache);
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -557,7 +557,7 @@ PHP_APCU_API zend_bool apc_cache_update_ttl(
 	entry = apc_cache_rlocked_find_nostat(cache, key, t);
 	if (!entry) {
 		apc_cache_runlock(cache);
-		return 0
+		return 0;
 	}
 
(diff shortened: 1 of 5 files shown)


apc_cache.c
```diff
@@ -539,6 +539,34 @@ PHP_APCU_API zend_bool apc_cache_store(
 	return ret;
 } /* }}} */
 
+/* {{{ apc_cache_set_ttl */
+PHP_APCU_API zend_bool apc_cache_update_ttl(
+		apc_cache_t* cache, zend_string *key, const int32_t ttl) {
```
nikic, line 544:
```suggestion
		apc_cache_t *cache, zend_string *key, int32_t ttl) {
```

php_apc.c
```diff
@@ -517,6 +517,26 @@ PHP_FUNCTION(apcu_store) {
...
+	zval *success = NULL;
+	time_t t;
+
+	ZEND_PARSE_PARAMETERS_START(1, 2)
+		Z_PARAM_STR(key)
+		Z_PARAM_OPTIONAL
+		Z_PARAM_LONG(ttl)
+	ZEND_PARSE_PARAMETERS_END();
+
+	t = apc_time();
```
nikic, line 534: Unused?

php_apc.c
```diff
@@ -517,6 +517,26 @@ PHP_FUNCTION(apcu_store) {
 }
 /* }}} */
 
+/* {{{ proto bool apc_cache_update_ttl(mixed key [, long ttl ])
+ */
+PHP_FUNCTION(apcu_set_ttl) {
+	zend_string *key;
+	zend_long ttl = 0L;
+	zval *success = NULL;
```
nikic, line 525: Unused?

php_apc.c
```diff
@@ -517,6 +517,26 @@ PHP_FUNCTION(apcu_store) {
...
+	time_t t;
+
+	ZEND_PARSE_PARAMETERS_START(1, 2)
+		Z_PARAM_STR(key)
+		Z_PARAM_OPTIONAL
+		Z_PARAM_LONG(ttl)
```
nikic, line 531: Why is ttl optional?

apc_cache.c
```diff
@@ -539,6 +539,34 @@ PHP_APCU_API zend_bool apc_cache_store(
...
+		return 0;
+	}
+
+	entry = apc_cache_rlocked_find_nostat(cache, key, t);
+	if (!entry) {
+		apc_cache_runlock(cache);
+		return 0;
+	}
+
+	entry->ctime = t;
```
nikic, line 563: Hm, should this really be setting ctime and not mtime (or neither)?

nikic (commented):
The feature sounds reasonable to me. This is currently missing a test though.

madmajestro:
@ignas2526
Since this PR is inactive and I also find this feature useful, I would like to ask if you mind if I open a new PR for this?

apc_cache.c
madmajestro: @nikic
I'm wondering if it wouldn't be generally better to change the behavior of apc_cache_entry_hard_expired() and apc_iterator_check_expiry() to check entries based on mtime instead of ctime.

In my opinion, it makes sense to calculate the TTL from the time of the last modification. Otherwise, you change something just before expiration, and then the entry is gone. That doesn't make sense to me.

ignas2526: So, my idea was to have a way to extend TTL of an entry. The only way to do it is to either change ctime or mtime as far as I understand. I don't think we should add new timestamp to track lifespan and make things complicated.

If I remember correctly, my idea was to do what apcu_add/apcu_store would do if it was called with the same values.

ignas2526:
> @ignas2526 Since this PR is inactive and I also find this feature useful, I would like to ask if you mind if I open a new PR for this?

Yea sure, you can take over if you have time. I think this PR is mostly done and the idea is outlined already. So you should have everything to wrap it up.

apc_cache.c
madmajestro: It is not my plan to introduce a new timestamp. But ctime should imho not be changed because it is the creation timestamp and not the modification timestamp. So my plan is to refresh mtime. But this currently makes setting ttl work not as expected as you would have to increase ttl with each ttl-update if you just plan to restart timer withe same ttl. The Solution is to change all expiery checks to mtime, which imho generally makes sense...

nikic: Using mtime sounds reasonable to me.

madmajestro:
@ignas2526
Thanks for your reply and your work, as it is mostly done!  I will then try to provide a PR for this as soon as i have time for this.

PR #520, opened 2024-09-02
Autotools: Refactor rt library check

petk:
This appends the -lrt only as needed to the LIBS (if apcu is built as static) or to the APCU_SHARED_LIBADD variable (when APCu is built as shared). Modern *nix systems have shared memory operations located in default libraries and explicit linking with -lrt is redundant.

~Edit: fixing this for the case when shm_open isn't found...~

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -221,7 +221,7 @@ if test "$PHP_APCU" != "no"; then
   AC_SEARCH_LIBS([shm_open], [rt])
   LIBS=$LIBS_SAVED
   AS_CASE([$ac_cv_search_shm_open], ["none required"|no], [],
-    [PHP_ADD_LIBRARY([rt], [], [APCU_SHARED_LIBADD])])
+    [PHP_EVAL_LIBLINE([$ac_cv_search_shm_open], [APCU_SHARED_LIBADD])])
 
   PHP_NEW_EXTENSION([apcu], [$apc_sources], [$ext_shared],, [$APCU_CFLAGS])
   PHP_SUBST(APCU_SHARED_LIBADD)
--- a/config.m4
+++ b/config.m4
@@ -217,7 +217,12 @@ if test "$PHP_APCU" != "no"; then
                  apc_iterator.c \
                  apc_persist.c"
 
-  PHP_CHECK_LIBRARY(rt, shm_open, [PHP_ADD_LIBRARY(rt,,APCU_SHARED_LIBADD)])
+  LIBS_SAVED=$LIBS; LIBS=
+  AC_SEARCH_LIBS([shm_open], [rt])
+  LIBS=$LIBS_SAVED
+  AS_VAR_IF([ac_cv_search_shm_open], ["none required"], [],
+    [PHP_ADD_LIBRARY([rt], [], [APCU_SHARED_LIBADD])])
+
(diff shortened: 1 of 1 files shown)


config.m4
```diff
@@ -217,7 +217,12 @@ if test "$PHP_APCU" != "no"; then
                  apc_iterator.c \
                  apc_persist.c"
 
-  PHP_CHECK_LIBRARY(rt, shm_open, [PHP_ADD_LIBRARY(rt,,APCU_SHARED_LIBADD)])
+  LIBS_SAVED=$LIBS; LIBS=
+  AC_SEARCH_LIBS([shm_open], [rt])
+  LIBS=$LIBS_SAVED
+  AS_CASE([$ac_cv_search_shm_open], ["none required"|no], [],
+    [PHP_EVAL_LIBLINE([$ac_cv_search_shm_open], [APCU_SHARED_LIBADD])])
```
nikic, line 224: Can we do this in action-if-found instead of a separate AS_CASE?

petk: I've rechecked this and I'm afraid that it can't be done inside the action-if-found argument. Because the rt library would get removed from the LIBS variable because of that `LIBS=$LIBS_SAVED` part when apcu would be built statically (inside php-src/ext, for example). For shared extension it would work, yes.

nikic: Wouldn't something like this work?
```
  LIBS_SAVED=$LIBS; LIBS=
  AC_SEARCH_LIBS([shm_open], [rt], [
    PHP_EVAL_LIBLINE([$LIBS], [APCU_SHARED_LIBADD])
  ])
  LIBS=$LIBS_SAVED
```
Or does LIBS only get modified afterwards?

petk: LIBS is modified before the action-if-found is executed.

```m4
AC_MSG_NOTICE([1: LIBS=$LIBS])
LIBS_SAVED=$LIBS; LIBS=
AC_SEARCH_LIBS([shm_open], [rt], [
  AC_MSG_NOTICE([2: LIBS=$LIBS])
  PHP_EVAL_LIBLINE([$LIBS], [APCU_SHARED_LIBADD])
  AC_MSG_NOTICE([3: LIBS=$LIBS])
])
AC_MSG_NOTICE([4: LIBS=$LIBS])
LIBS=$LIBS_SAVED
AC_MSG_NOTICE([5: LIBS=$LIBS])
```

Will produce something like this (when building APCu inside php-src and on some system (like some older Linux) that needs the rt library linked:

```
configure: 1: LIBS=-lm  -lxml2 -lsqlite3
checking for library containing shm_open... -lrt
configure: 2: LIBS=-lrt 
configure: 3: LIBS=-lrt  -lrt
configure: 4: LIBS=-lrt  -lrt
configure: 5: LIBS=-lm  -lxml2 -lsqlite3
```

There will be two -lrt added due to action-if-found and then LIBS will be reset afterwards. When building extension as shared, this would work.

petk: And, yes. These checks get uglier and uglier. I completely agree. PHP would here need something like PHP_SEARCH_LIBRARY besides the PHP_CHECK_LIBRARY. But adding this macro to php-src will be then available in apcu realistically somewhere in 5-10 years :D

PR #524, opened 2024-09-18
Added a composer.json to enable support for PIE

asgrim:
Adds support for [PIE](https://github.com/php/pie).

Once merged, a maintainer would need to add this repo's URL to https://packagist.org/packages/submit :)

More info/documentation can be read on: https://github.com/php/pie/blob/main/docs/extension-maintainers.md

PR #525, opened 2024-09-19
Update vendor name to apcu for PIE

asgrim:
As per discussion https://github.com/krakjoe/apcu/pull/524#issuecomment-2359423616


nikic:
Thanks, I was able to add it to packagist now: https://packagist.org/packages/apcu/apcu Unfortunately I don't have admin access to this repo, so I can't set up the auto-update hook.

PR #528, opened 2024-09-29
Use caching for Windows CI

cmb69:
Looks like a nice performance improvement to me. And this way, AppVeyor can be disabled. I would drop some Windows jobs, though; maybe run only ZTS?


cmb69:
Just updated setup-php-sdk to v0.10.

PR #533, opened 2025-02-27
Fix apc_cache_default_expunge() to always remove expired entries

madmajestro:
When apc_cache_default_expunge() is called, it now always cleans up expired entries and doesn't skip cleanup anymore, as it is only called when there is not enough memory available to store a cache entry and cleanup is really needed. The skip can happen because the function apc_sma_get_avail_mem() doesn't take into account that a block of contiguous memory is needed. In situations with huge cache fragmentation, the old behavior can cause apcu to not accept new entries anymore which don't fit into a fragment of available memory. This causes php-functions like apcu_store() to not being able to insert bigger entries, until the cache has filled up enough with small entries to trigger the cleanup. This commit fixes the issue and mostly implements a behavior that should have already been implemented previously according to the header file. This also applies to the smart configuration setting that is now used to tune the probability of a full cache wipe.

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -735,8 +735,7 @@ PHP_APCU_API void apc_cache_clear(apc_cache_t* cache)
 PHP_APCU_API void apc_cache_default_expunge(apc_cache_t* cache, size_t size)
 {
 	time_t t;
-	size_t suitable = 0L;
-	size_t available = 0L;
+	size_t i;
 
 	if (!cache) {
 		return;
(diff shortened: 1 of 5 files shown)


madmajestro:
... and can you give me a hint / link on how I can publish documentation for the "smart" configuration parameter (e.g. on php.net)?

madmajestro:
I just saw that i didn't explain the exact reason why the faulty skip can happen: The skip can happen because the function apc_sma_get_avail_mem() doesn't take into account that a block of contiguous memory is needed. But it doesn't make sense to fix apc_sma_get_avail_mem(), because it is used in other code paths and the needed check is already done in the sma layer in a correct way (which leads to the call oft the expunge function). So it is a buggy double-check, which is removed by this pr. The commit text has been slightly adjusted to reflect this.

madmajestro:
I added two tests that demonstrate the unexpected behavior on the master branch (which is solved by this pr). Any chance to get a review for this? Please let me know if I'm wrong or if I can do better.

mszabo-wikia:
@madmajestro Thanks for this! Regarding your question about updating docs, they live at https://github.com/php/doc-en/tree/master/reference/apcu.

madmajestro:
superseded by #536 and #537

PR #534, opened 2025-03-01
Fix required memory size passed to expunge()

madmajestro:
The size passed to expunge() must match the size passed to find_block().

### DIFF
--- a/apc_sma.c
+++ b/apc_sma.c
@@ -449,7 +449,7 @@ restart:
 
 	/* Expunge cache in hope of freeing up memory, but only once */
 	if (!nuked) {
-		sma->expunge(*sma->data, n+fragment);
+		sma->expunge(*sma->data, n);
 		nuked = 1;
 		goto restart;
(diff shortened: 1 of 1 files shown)


apc_sma.c
```diff
@@ -449,7 +449,7 @@ PHP_APCU_API void *apc_sma_malloc_ex(apc_sma_t *sma, size_t n, size_t *allocated
 
 	/* Expunge cache in hope of freeing up memory, but only once */
 	if (!nuked) {
-		sma->expunge(*sma->data, n+fragment);
+		sma->expunge(*sma->data, ALIGNWORD(n + ALIGNWORD(sizeof(struct block_t))));
```
nikic, line 452: I don't think we should be passing the real size to the expunge hook. Instead apc_sma_get_avail_size should be computing the real size for its check.

madmajestro: yes, I absolutely agree. Fixed.

PR #536, opened 2025-03-07
Cleanup of entries should not be skipped when using default settings

madmajestro:
By default (apc.smart = 0), apc_cache_default_expunge() should not skip cleanup of old entries. This may cause insertion of new entries to fail unnecessarily, which is unexpected behavior.

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -735,8 +735,7 @@ PHP_APCU_API void apc_cache_clear(apc_cache_t* cache)
 PHP_APCU_API void apc_cache_default_expunge(apc_cache_t* cache, size_t size)
 {
 	time_t t;
-	size_t suitable = 0L;
-	size_t available = 0L;
+	size_t i;
 
 	if (!cache) {
 		return;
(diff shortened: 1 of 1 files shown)


LionsAd (approved):
FWIW, here is a peer-review: Looks great to me!

And is fully opt-out via `apc.smart=1`.

madmajestro:
@LionsAd
Thanks for your review and your approval!

Just to be sure, since you mentioned `apc.smart = 1` as a fully opt-out...
Using `apc.smart = 1` you will get the same behavior as before with `apc.smart = 1`. But that's not what `apc.smart = 0` (default) did before. Therefore, strictly speaking, it not really a fully opt-out.

The main purpose of this PR is to remove the skip that can occur, when  `apc.smart = 0` AND `apc_sma_get_avail_mem(cache->sma) >= (cache->sma->size / 2)`.

I think this skip was intended as some kind of optimization, but when the expunge() hook is called, in the vast majority of cases this skip does not occur. So it optimizes almost nothing. But when it occurs, it causes unexpected behavior (PHP functions like apcu_store() are not able to store the new entry). That's why I think it makes sense to remove this skip.

LionsAd:
Yes, that is what I meant that with the old smart = 1 behavior, you have the same behavior.

And I agree that it’s better to purge some items from time to time than to completely run out of memory.

nikic:
To check my understanding, this is only about the case where you're trying to store something that's larger than half the entire cache size?

apc_cache.c
```diff
@@ -751,43 +750,38 @@ PHP_APCU_API void apc_cache_default_expunge(apc_cache_t* cache, size_t size)
...
-
-		/* look for junk */
-		for (i = 0; i < cache->nslots; i++) {
-			apc_cache_entry_t **entry = &cache->slots[i];
-			while (*entry) {
-				if (apc_cache_entry_expired(cache, *entry, t)) {
-					apc_cache_wlocked_remove_entry(cache, entry);
-					continue;
-				}
+	if (cache->smart > 0L && apc_sma_get_avail_mem(cache->sma) >= (size_t) (cache->smart * size)) {
```
nikic, line 757: Does this even do anything useful? We're only going to call expunge if there is not enough space for size, right?

nikic: Ah no, this is checking for total free size, not available single block. But then doesn't setting smart > 0 still result in expunge being skipped sometimes even though it may have been possible to free up consecutive space?

nikic:
> To check my understanding, this is only about the case where you're trying to store something that's larger than half the entire cache size?

Okay, I think I misunderstood, and what this is trying to fix is a high-fragmentation case where half the cache is empty, but not enough space for a consecutive allocation?

madmajestro:
Exactly!

In #533, I solved both problems (no skipping in high fragmentation situations + usefulness of smart setting), but it's a bit verbose...

Therefore, if this PR (#536) is merged, I'd like to provide a solution for the smart setting in another PR. I will close #533 then.

apc_cache.c
madmajestro: Yes, it does and I'd like to change how smart works to address this issue as well. I made a suggestion how smart could be changed in #533. Then I opened this pr (#536) because i wasn't sure if it is a good idea to change the behavior of smart. But finally i think we should do it, because the current implementation seems to be not very useful and will cause unexpected behavior.

If you think it makes sense to change the behavior of smart, I could make the entire change (including smart) here or open a new PR for it. Which do you prefer?

madmajestro:
Friendly ping.

nikic:
I went ahead and merged this for now. I'm not particularly happy with the smart behavior here, but TBH I don't know what the correct behavior is even supposed to be there.

I'm also somewhat concerned about fully expunging the cache in high fragmentation scenarios where there is still a lot of free space. I think that ideally, we'd make the persistence representation position-independent (via pointer unswizzling), which would then allow relocating the cache entries to reduce fragmentation. This would be a large change though.

Another option to consider would be to change the allocator to reduce fragmentation (via size binning?) The current implementation is naive.

PR #537, opened 2025-03-17
Change behavior of smart configuration setting

madmajestro:
The smart configuration setting can now be used to increase the chance of full cache wipes during periods of high memory pressure. The idea is to use smart values > 1 to avoid performance problems when the expunge() hook is executed too often. The previous behavior was changed because it could unpredictably cause the insertion of new cache entries to fail, making the smart setting nearly useless.


madmajestro:
This is a follow-up of #536.

This is my suggestion for how the smart setting should work / could be improved.

I suspect that the smart setting should have worked this way before.

nikic (approved):
Makes sense to me.

PR #539, opened 2025-04-01
Fix computation of minimum fragment size

madmajestro:
The minimum fragment size previously used by the sma layer was too small. This could lead to the unnecessary creation of small free blocks that could not be used for any allocation. To keep separation of layers, the minimum size is now passed to apc_sma_init().

### DIFF
--- a/apc_cache.h
+++ b/apc_cache.h
@@ -299,6 +299,8 @@ static inline void apc_cache_runlock(apc_cache_t *cache) {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


madmajestro:
The fragment size previously used by the sma layer to decide whether to split a block was 80 bytes `(MINBLOCKSIZE + Fragment)`, even though the smallest possible cache entry requires a contiguous block of 160 bytes (using `apcu_store("", 1)`).

This could result in small free blocks that could never be allocated and had to be checked with each call to find_block().

nikic (approved):
Nice!

PR #540, opened 2025-04-02
Remove unused code disabled by define APC_MEMPROTECT

madmajestro:


madmajestro:
As far as I can tell, this code hasn't been needed for many years (perhaps since APCU was derived from APC). Therefore, I don't see any further use for it. Furthermore, there's no way to compile this code in the build system, so I believe it can be safely removed.

PR #541, opened 2025-04-04
Cleanup mmap and shm implementation

madmajestro:
Better separation of the sma layer and the mmap / shm implementation.


madmajestro:
This also makes the mmap and shm implementations independent of `apc_segment_t`, which could be useful as a first step toward removing support for multiple shared memory segments in the future.

PR #544, opened 2025-04-08
Unify shmaddr and header in sma layer

madmajestro:
The variables shmaddr and header were often used simultaneously in the same context, although both always pointed to the same memory address. This was because the BLOCKAT / OFFSET macros expected shmaddr to be present and header was used to access data of the header struct. Now the variable smaheader is used for both.


madmajestro:
I chose the name smaheader (instead of header) to better distinguish this header from the header of the individual data blocks.

PR #545, opened 2025-04-09
Avoid different notations to calculate the size of block_t

madmajestro:


madmajestro:
The notation `sizeof(struct xxx)` is not used anywhere else in the project.

PR #547, opened 2025-04-09
Fix size of initial empty block in sma layer

madmajestro:
The sum of the size of all free blocks is now in sync with smaheader->avail. Additionally, this fixes the position of the last (0-sized) block. Before this change, ALIGNWORD(sizeof(block_t)) bytes remained free at the end of the shared memory.

### DIFF
--- a/apc_sma.c
+++ b/apc_sma.c
@@ -310,7 +310,7 @@ PHP_APCU_API void apc_sma_init(apc_sma_t* sma, void** data, apc_sma_expunge_f ex
 		SET_CANARY(first);
 
 		empty = BLOCKAT(first->fnext);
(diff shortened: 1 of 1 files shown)


nikic (commented):
Any idea why this was here in the first place?

madmajestro:
I traced this back to a commit in the `php/pecl-caching-apc` repository from 2008-08-13 (bb174256). That's when the terminating 0-sized block and prev references were introduced for faster deallocation. I think this was an oversight in that commit and has been incorrect since then.

madmajestro:
I've reviewed the SMA logic and haven't found any reason why this might have been necessary. Please let me know if you have any concerns.

PR #551, opened 2025-04-21
Refactor sma to use pointers instead of offsets

madmajestro:
The block_t struct now uses pointers instead of offset values to refer to other blocks. This simplifies code and should also slightly improve performance by removing unnecessary pointer computations.

### DIFF
--- a/TECHNOTES.txt
+++ b/TECHNOTES.txt
@@ -73,14 +73,14 @@ form of a quick-start guide to start hacking on APCu.
      | header | 0-size |        free-block        | 0-size |
      +--------+--------+-----------------------------------+
 
-   The blocks are just a simple offset-based doubly linked list (so no pointers):
+   The blocks are just a simple doubly linked list:
 
      typedef struct block_t block_t;
      struct block_t {
-         size_t size;       /* size of this block */
-         size_t prev_size;  /* size of sequentially previous block, 0 if prev is allocated */
-         size_t fnext;      /* offset in segment of next free block */
-         size_t fprev;      /* offset in segment of prev free block */
+         size_t   size;     /* size of this block */
+         block_t *fnext;    /* pointer to next free block in segment */
+         block_t *fprev;    /* pointer to prev free block in segment */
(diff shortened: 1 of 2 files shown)


madmajestro:
I see no advantages, only disadvantages, in using offsets compared to pointers in the sma logic.
Therefore, I refactored sma to use pointers in block_t.

nikic:
The advantage of using offsets is that it allows mapping sma at different addresses. This currently doesn't help because the cache itself is not relocatable. But making it relocatable may be worthwhile to allow compacting fragmented caches without expunging. And if both the cache and sma are relocatable, then apcu could actually support Windows again.

madmajestro:
> The advantage of using offsets is that it allows mapping sma at different addresses. This currently doesn't help because the cache itself is not relocatable. But making it relocatable may be worthwhile to allow compacting fragmented caches without expunging. And if both the cache and sma are relocatable, then apcu could actually support Windows again.

I wasn't aware that apcu doesn't currently support Windows. If it was possible to get the same base address for the shared memory before, what did cause the breakage?

For defragmentation my intention was to write a function that modifies all pointers in the SHM representation of an entry to a different absolute position. I thought this approach might keep the overall code a bit cleaner compared to offsets.

However, if there is no other way to support Windows again, and if the plan is to allow other PHP CLI processes to map to an existing SHM, this change makes no sense and we need offsets in the cache layer instead.

madmajestro:
> I want to continue working on making it possible for commandline to attach to an existing SMA / SHM later (I came close).
> 
> I think pointers should not change so this should still work, because it's a fixed shared memory region, but I also think it would be great to have offsets to make it re-locatable and potentially growable.
> 
> Thanks for all your work on cleaning the SMA layer up!

This doesn't work without offsets in the cache layer. I can take a look at the offset implementation in the cache layer.

madmajestro:
In any case, I think it makes sense to drop support for multiple shared memory segments in the sma layer as a first step, as this will simplify the implementation of offsets in the cache layer. Furthermore, I can't think of any system where multiple segments are still necessary or useful nowadays. I will provide a PR for this.

PR #552, opened 2025-04-24
Drop support for multiple shared memory segments

madmajestro:
The support for multiple shared memory segments is not necessary these days, as systems that suffered from shared memory size limitations now use mmap(), which is unaffected by such limitations. In addition, this simplifies the sma layer and makes it easier to refactor the cache layer so that the cache is relocatable.


madmajestro:
As mentioned in #551, i propose to drop the support for multiple shared memory segments. I didn't change the array returned by `apcu_sma_info()` to ensure compatibility with existing PHP scripts. This can be changed later if necessary.

PR #553, opened 2025-04-25
Use mtime instead of ctime for hard expiration checks

madmajestro:
Using mtime instead of ctime for expiration checks causes an entry's TTL timer to restart when the entry is changed. This prevents entries from expiring shortly after a change, which makes no sense.

### DIFF
--- a/TECHNOTES.txt
+++ b/TECHNOTES.txt
@@ -265,7 +265,7 @@ form of a quick-start guide to start hacking on APCu.
         zend_long ref_count;     /* the reference count of this entry */
         zend_long nhits;         /* number of hits to this entry */
         time_t ctime;            /* time entry was initialized */
-        time_t mtime;            /* the mtime of this cached entry */
+        time_t mtime;            /* time entry was last modified */
         time_t dtime;            /* time entry was removed from cache */
         time_t atime;            /* time entry was last accessed */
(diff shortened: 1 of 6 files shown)


madmajestro:
This also allows a cleaner implementation of functions like apcu_update_ttl(), as discussed in #507. The tests 021 and 022 did fail after switching to mtime. Therefore I had to change them.

madmajestro:
I forgot to mention that I renamed mtime in apc_cache_slam_key_t to ctime, because it's a creation time. The terms shouldn't be mixed up.

nikic:
Hm, looking at this, I suspect this is going to break code. For example, I found this one from a search of "apcu rate limit": https://github.com/nikolaposa/rate-limit/blob/master/src/ApcuRateLimiter.php

Basically, the way it currently works, is that you can do crude rate limiting by creating a key with a certain TTL and then increment it until it exceeds some limit.

After this change, the TTL would always get extended on increment, so you would end up enforcing the limit over a much longer time period than intended.

madmajestro:
Yes, you're right. We shouldn't switch the expiration logic from ctime to mtime and I need to find another solution for the implementation of `apcu_set_ttl()`. Therefore, I'm closing this PR.

PR #554, opened 2025-04-30
Make the key of a cache entry relocatable

madmajestro:
The key is now implemented as the last part of apc_cache_entry_t, which is another step toward making entries relocatable in the future. Furthermore, this reduces the size of an entry by the size of a pointer and removes one memcpy() of apc_cache_entry_t during insertion.

### DIFF
--- a/apc_persist.c
+++ b/apc_persist.c
@@ -86,7 +86,7 @@ static inline void apc_persist_copy_zval(apc_persist_context_t *ctxt, zval *zv) 
 	apc_persist_copy_zval_impl(ctxt, zv);
 }
 
-void apc_persist_init_context(apc_persist_context_t *ctxt, apc_serializer_t *serializer) {
+static void apc_persist_init_context(apc_persist_context_t *ctxt, apc_serializer_t *serializer) {
 	ctxt->serializer = serializer;
 	ctxt->size = 0;
 	ctxt->memoization_needed = 0;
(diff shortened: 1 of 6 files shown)


madmajestro:
The next step would be to make the `zval` of `apc_cache_entry_t` relocatable. I will provide a pr for this as well.

apc_persist.c
```diff
@@ -435,40 +435,53 @@ static void apc_persist_copy_zval_impl(apc_persist_context_t *ctxt, zval *zv) {
 	}
 }
 
-static apc_cache_entry_t *apc_persist_copy(
-		apc_persist_context_t *ctxt, const apc_cache_entry_t *orig_entry) {
-	apc_cache_entry_t *entry = COPY(orig_entry, sizeof(apc_cache_entry_t));
-	entry->key = apc_persist_copy_zstr_no_add(ctxt, entry->key);
+static apc_cache_entry_t * apc_persist_create_entry(
```
nikic, line 438:
```suggestion
static apc_cache_entry_t *apc_persist_create_entry(
```

apc_persist.c
```diff
@@ -435,40 +435,53 @@ static void apc_persist_copy_zval_impl(apc_persist_context_t *ctxt, zval *zv) {
...
-		apc_sma_t *sma, apc_serializer_t *serializer, const apc_cache_entry_t *orig_entry) {
+		apc_sma_t *sma, apc_serializer_t *serializer, zend_string *key, const zval *val) {
 	apc_persist_context_t ctxt;
 	apc_cache_entry_t *entry;
 
 	apc_persist_init_context(&ctxt, serializer);
 
 	/* The top-level value should never be a reference */
-	ZEND_ASSERT(Z_TYPE(orig_entry->val) != IS_REFERENCE);
+	ZEND_ASSERT(Z_TYPE(*val) != IS_REFERENCE);
```
nikic, line 467:
```suggestion
	ZEND_ASSERT(Z_TYPE_P(val) != IS_REFERENCE);
```
Here and elsewhere

madmajestro:
Fixed.

PR #556, opened 2025-05-02
Make cache entries relocatable

madmajestro:
This makes the persistence representation of cache entries in shm position-independent by converting all pointers in an entry's zval value to offsets relative to the entry's starting address. During persistence, all (process local) pointers are converted and stored as offsets in shm. During unpersistence, the (process local) shm starting address of the entry is added to each offset before data access. This is necessary to make the whole cache relocatable or to develop defragmentation.

### DIFF
--- a/apc_persist.c
+++ b/apc_persist.c
@@ -53,7 +53,7 @@ typedef struct _apc_persist_context_t {
 	/* Serialized object/array string, in case there can only be one */
 	unsigned char *serialized_str;
 	size_t serialized_str_len;
-	/* Whole SMA allocation */
+	/* Address (process local) of entry in shm / Whole SMA allocation */
 	char *alloc;
 	/* Current position in allocation */
 	char *alloc_cur;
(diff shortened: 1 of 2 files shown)


madmajestro:
I plan to provide pull request for defragmentation / make the entire cache relocatable as well.

apc_persist.c
```diff
@@ -41,6 +41,15 @@
  * PERSIST: Copy from request memory to SHM.
  */
 
+/* The macros TO_OFF and TO_PTR are used for convenience throughout this module.
+ * Both assume the presence of a variable ctxt that points to an
+ * apc_persist_context_t or apc_unpersist_context_t structure.
+ * TO_OFF converts a (process local) pointer to an offset
+ * relative to the beginning of a cache entry.
+ * TO_PTR reverts the offset to a (process local) pointer. */
+#define TO_OFF(ptr) ((void *)((char *)ptr - (size_t)ctxt->alloc))
```
nikic, line 50: It's best to do the arithmetic in uintptr_t to avoid UB.

nikic: Can we add an assertion that ptr is >= alloc and < alloc + size? I suspect that you are currently using TO_OFF on pointers outside SMA as well.

madmajestro: I fixed the arithmetic to use uintptr_t and added a scratch commit for debugging TO_OFF() which i would like to drop before the pr is merged.

I haven't found an access using TO_OFF() to memory outside sma.

Hmm, do you suspect `ht->arPacked = TO_OFF(ht->arPacked);` or `ht->arData = TO_OFF(ht->arData);` in `apc_persist_copy_ht()`?

These are imho ok because of this code before:
`HT_SET_DATA_ADDR(ht, COPY(HT_GET_DATA_ADDR(ht), HT_PACKED_USED_SIZE(ht)));`

HT_GET_DATA_ADDR(ht) gets an address outside from sma, but the COPY macro copies it to shm and gives back the target shm address, which is then set by HT_SET_DATA_ADDR to `ht->arData`...

madmajestro: Or did I misunderstand and you meant to permanently include the assertion using static inline helper functions?

nikic: I think it's a good idea to permanently include the assertion (don't need the separate warning though).

madmajestro: Fixed. I added assertions for both conversions and ran some tests with assert() and ZEND_ASSERT().
Using assert() produces the same assembly code (non-a debug build) as the macro-only version.
ZEND_ASSERT(), however, does not (maybe because it turns into an assume()). Is ZEND_ASSERT() a good choice here?

nikic: Should probably use plain assert() in that case. Generating an assumption here can only make things worse, not better...

madmajestro: Fixed.

nikic (approved):
This was simpler than I expected...

madmajestro:
Yes, after I had the idea of ​​how to do it, I thought the same thing...

PR #557, opened 2025-05-06
Make cache relocatable

madmajestro:
This makes the cache position-independent by converting all pointers to entries into offsets relative to the starting address of the cache header in the shared memory segment. This is a further step to enable the use of APCU with processes that access the shared memory segment via different starting addresses.

### DIFF
--- a/apc_cache.h
+++ b/apc_cache.h
@@ -50,7 +50,7 @@ struct apc_cache_slam_key_t {
 typedef struct apc_cache_entry_t apc_cache_entry_t;
 struct apc_cache_entry_t {
 	zval val;                /* the zval copied at store time */
-	apc_cache_entry_t *next; /* next entry in linked list */
+	uintptr_t next;          /* offset in shm of next entry in linked list */
 	zend_long ttl;           /* the ttl on this specific entry */
 	zend_long ref_count;     /* the reference count of this entry */
 	zend_long nhits;         /* number of hits to this entry */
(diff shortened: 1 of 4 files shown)


madmajestro:
To allow processes to connect to an existing cache, some logic must be implemented to detect whether the process is connecting to an already initialized cache to skipp an additional header initialization that would break the cache. This is not part of this PR.

apc_cache.h
```diff
@@ -300,6 +301,12 @@ static inline void apc_cache_runlock(apc_cache_t *cache) {
 /* APC_ENTRY_SIZE takes into account the trailing key-string + terminating 0-byte */
 #define APC_ENTRY_SIZE(key_len) (ZEND_MM_ALIGNED_SIZE(XtOffsetOf(apc_cache_entry_t, key.val) + key_len + 1))
 
+/* ENTRYAT and ENTRYOF are used to convert between offsets and pointers to cache entries.
+ * Both expect the presence of apc_user_cache->shmaddr that points to the beginning
+ * of the shared memory segment. */
+#define ENTRYAT(offset) ((apc_cache_entry_t *)((uintptr_t)apc_user_cache->shmaddr + (uintptr_t)offset))
```
nikic, line 307: Can we use cache instead of apc_user_cache in these?

apc_cache.h
```diff
@@ -75,13 +75,14 @@ typedef struct _apc_cache_header_t {
 	zend_long mem_size;             /* used */
 	time_t stime;                   /* start time */
 	apc_cache_slam_key_t lastkey;   /* last key inserted (not necessarily without error) */
-	apc_cache_entry_t *gc;          /* gc list */
+	uintptr_t gc;                   /* offset in shm to the first entry of gc list */
 } apc_cache_header_t; /* }}} */
 
 /* {{{ struct definition: apc_cache_t */
 typedef struct _apc_cache_t {
+	void *shmaddr;                /* address (process local) of the shared memory segment */
```
nikic, line 83: Thoughts about making the offsets relative to header rather than shmaddr?

madmajestro: done.

apc_cache.h
madmajestro: Done. I didn't like that either. I originally did it that way because the cache variable wasn't present in the iterator logic, and I didn't want to change the iterator logic too much in this pr. But I think it's better the way it is now.

nikic (approved):

PR #558, opened 2025-05-08
Compute the number of hash slots instead of using a fix default value

### LINKED ISSUE #344: Increase default entries_hint to 16k / or use auto detection / ensure apc.php suggests doing so
### Motivation

With just 4k entries in the hash table, it means that more than 4k entries will automatically lead to long lists within the hash buckets, which decreases performance a lot as list traversal is obviously slow.

Also it is currently not easily seeable from apc.php how many entries are present in the cache -- just the fragment information is there visible.

I think for most applications 16k is a better default, at least if memory > 64M.

So maybe a good idea would be to scale the hash table size with the memory_size by default and give a good visible hint that `entries >> entries_hint` and hence reduced performance.

### Implementation

It feels best to scale entries_hint dynamically with cache size:

- 32M => 4k
- 64M => 8k
- 128M => 16k
- 256M => 32k

etc.

given that overall potential memory waste would not change, that feels like a good idea.

(This was another suggestion from #333 that I had.)

madmajestro:
The new default behavior is to use 512 slots per MB of shared memory, which keeps the hash table size < 0.4% of shm, which seems negligible. This should improve the default behavior, as it improves apcu's performance with many entries, and the hash table size now scales with the size of shared memory. Previously, a fixed value of 4096 at 32 MB shm was used as default, which corresponds to 128 slots per MB.

Closes #344

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -288,8 +288,8 @@ PHP_APCU_API apc_cache_t* apc_cache_create(apc_sma_t* sma, apc_serializer_t* ser
 	zend_long cache_size;
 	size_t nslots;
 
-	/* calculate number of slots */
-	nslots = make_prime(size_hint > 0 ? size_hint : 2000);
+	/* calculate number of slots. Default: 512 slots per MB of shared memory */
+	nslots = make_prime(size_hint > 0 ? (size_t)size_hint : sma->size / 2048);
 
 	/* allocate pointer by normal means */
 	cache = pemalloc(sizeof(apc_cache_t), 1);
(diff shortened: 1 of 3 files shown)


madmajestro:
I propose to compute the number of hash slots by default so that, equivalent to 32MB, 16k hash slots are used. The size of the hash table is still negligible in relation to the total cache size, but in most cases it leads to good default behavior, eliminating the need for further tuning. The biggest problem I see is that many people who configure apcu have no idea what values ​​for entries_hint make sense. So we'll take this decision away from them and set a sensible default.

madmajestro:
Just a thought:
Currently, the minimum entry size is 152 bytes. This means that in the worst case, there can be approximately 6898.5 entries per MB. With the current default, this means an average of approximately 54 entries per slot, and it gets worse if you increase the memory without adjusting entries_hint.

With the proposed new default behavior, we would always have approximately 13.5 entries per slot in the worst case.

nikic (approved):

PR #559, opened 2025-05-08
Add support for hugetlb pages on Linux

KitanoTom:
Enable huge pages on Linux by calling `mmap()` with the `MAP_HUGETLB` flag.
Using huge pages can improve performance when working with large amounts of shared memory.

This PR adds the `apc.mmap_hugetlb_mode` setting, which accepts the following values:
`none` – Disable HugeTLB (current behavior).
`default` – Pass `MAP_HUGETLB` and let the kernel choose the default huge-page size.
`16kb`~`16gb` – Pass `MAP_HUGETLB | MAP_HUGE_xx` and explicitly select a huge-page size.

---

In the final version, the `apc.mmap_hugepage_size` parameter was added instead of `apc.mmap_hugetlb_mode` to specify the page size.
The default is `0`, which is the same behavior as before.

The page size must be a power of 2, and `apc.shm_size` must be a multiple of the page size.

### DIFF
--- a/tests/apc_mmap_hugepage_002.phpt
+++ b/tests/apc_mmap_hugepage_002.phpt
@@ -10,6 +10,9 @@ if (ini_get('apc.mmap_hugepage_size') === false) die("skip mmap is not used");
 // currently only support Linux
 if (PHP_OS != "Linux") die("skip only on Linux");
 
+// does the kernel support MAP_HUGETLB?
+if (version_compare(preg_replace('/[^0-9.].*$/', '', php_uname('r')), '2.6.32', '<')) die("skip kernel does not support MAP_HUGETLB");
+
 // check reserved hugepage
 $hp = @file_get_contents('/proc/sys/vm/nr_hugepages');
 if ($hp === false || !((int)trim($hp))) die("skip hugepages are currently unavailable on this system");
--- a/tests/apc_mmap_hugepage_003.phpt
+++ b/tests/apc_mmap_hugepage_003.phpt
@@ -10,6 +10,9 @@ if (ini_get('apc.mmap_hugepage_size') === false) die("skip mmap is not used");
 // currently only support Linux
 if (PHP_OS != "Linux") die("skip only on Linux");
 
+// does the kernel support MAP_HUGETLB?
+if (version_compare(preg_replace('/[^0-9.].*$/', '', php_uname('r')), '2.6.32', '<')) die("skip kernel does not support MAP_HUGETLB");
+
 // check reserved hugepage
 $hp = @file_get_contents('/proc/sys/vm/nr_hugepages');
 if ($hp === false || !((int)trim($hp))) die("skip hugepages are currently unavailable on this system");
--- a/tests/apc_mmap_hugepage_004.phpt
+++ b/tests/apc_mmap_hugepage_004.phpt
@@ -10,6 +10,9 @@ if (ini_get('apc.mmap_hugepage_size') === false) die("skip mmap is not used");
 // currently only support Linux
 if (PHP_OS != "Linux") die("skip only on Linux");
 
+// does the kernel support MAP_HUGETLB?
+if (version_compare(preg_replace('/[^0-9.].*$/', '', php_uname('r')), '2.6.32', '<')) die("skip kernel does not support MAP_HUGETLB");
+
 // check reserved hugepage
 $hp = @file_get_contents('/proc/sys/vm/nr_hugepages');
 if ($hp === false || !((int)trim($hp))) die("skip hugepages are currently unavailable on this system");
--- a/tests/apc_mmap_hugepage_005.phpt
+++ b/tests/apc_mmap_hugepage_005.phpt
@@ -10,6 +10,9 @@ if (ini_get('apc.mmap_hugepage_size') === false) die("skip mmap is not used");
 // currently only support Linux
 if (PHP_OS != "Linux") die("skip only on Linux");
 
+// does the kernel support MAP_HUGETLB?
+if (version_compare(preg_replace('/[^0-9.].*$/', '', php_uname('r')), '2.6.32', '<')) die("skip kernel does not support MAP_HUGETLB");
+
 // check reserved hugepage
 $hp = @file_get_contents('/proc/sys/vm/nr_hugepages');
 if ($hp === false || !((int)trim($hp))) die("skip hugepages are currently unavailable on this system");
--- a/tests/apc_mmap_hugepage_006.phpt
+++ b/tests/apc_mmap_hugepage_006.phpt
@@ -10,6 +10,9 @@ if (ini_get('apc.mmap_hugepage_size') === false) die("skip mmap is not used");
 // currently only support Linux
 if (PHP_OS != "Linux") die("skip only on Linux");
 
+// does the kernel support MAP_HUGETLB?
+if (version_compare(preg_replace('/[^0-9.].*$/', '', php_uname('r')), '2.6.32', '<')) die("skip kernel does not support MAP_HUGETLB");
+
 // check reserved hugepage
 $hp = @file_get_contents('/proc/sys/vm/nr_hugepages');
 if ($hp === false || !((int)trim($hp))) die("skip hugepages are currently unavailable on this system");


KitanoTom:
I’ve updated the error message for mmap failures.

apc_mmap.c
```diff
@@ -51,7 +55,99 @@
...
+# ifndef MAP_HUGETLB
+	flags = 0;
+	apc_warning("MAP_HUGETLB is not defined on this system");
+# else
+	flags = MAP_HUGETLB;
+	if (!strcasecmp(mmap_hugetlb_mode, "default")) {
+		return flags;
+	}
+#  ifdef MAP_HUGE_16KB
+	if (!strcasecmp(mmap_hugetlb_mode, "16kb")) {
```
nikic, line 78: Instead of listing all of these individually, can we make use of MAP_HUGE_SHIFT instead?

tests/apc_hugetlb.phpt
```diff
@@ -0,0 +1,20 @@
+--TEST--
+Check apc.mmap_hugetlb_mode parameter
+--SKIPIF--
+<?php
+require_once(dirname(__FILE__) . '/skipif.inc'); 
+if (PHP_OS != "Linux") die("skip only on Linux");
```
nikic, line 6: Is this test going to work on a system with huge pages disabled (without throwing any errors?)

madmajestro:
Hm, my thoughts to this:
We already call `madvise(shmaddr, size, MADV_HUGEPAGE);` after calling mmap(). So I wonder if explicitly using HugePages during mmap() provides a significant benefit.

@KitanoTom:
To see the effect of MADV_HUGEPAGE, you need to ensure the following:
- `/sys/kernel/mm/transparent_hugepage/enabled` should be set to `allways` or `madvise` (my operating system's default is `allways`).
- `/sys/kernel/mm/transparent_hugepage/shmem_enabled` should not be disabled. I recommend setting it to `advise` if it is set to `never` (`never` is my operating system's default).

Once this is set up and you've restarted PHP, you can use `grep HugePage /proc/meminfo` to see if it has any effect (ShmemHugePages should grow while the cache fills up).

Imho, this should have almost the same effect as an explicit configuration of HugePages, and the configuration should be less complex because you don't need to allocate HugePages in the correct amount before starting php from the OS.

Please let me know if I am wrong or if I have missed something.

KitanoTom:
@nikic 
Thank you for your review.

I updated the implementation to use `MAP_HUGE_SHIFT`, and fixed test case.
In addition, the configuration parameter has been renamed from `apc.mmap_hugetlb_mode` to `apc.mmap_hugetlb_page_size`.

@madmajestro 
Thank you for your feedback.

For most standard use cases, I also feel that Transparent Huge Pages (THP) are more than adequate.
However, in the products my organization builds we sometimes need to reserve anywhere from several to a few-dozen GB of shared memory. In those situations it’s more convenient to use HugeTLB Pages, which lets us choose larger page sizes (e.g., 1GB pages) and tune things more precisely.
(My understanding is that THP tops out at 2MB pages; please correct me if I’m mistaken.)

madmajestro:
Maybe you are right and it's worth it in special use cases.

apc_mmap.c
```diff
@@ -35,6 +35,14 @@
...
 #include <sys/mman.h>
 
+#if defined(__linux__)
+# ifdef __has_include
+#  if __has_include(<linux/mman.h>)
+#   include <linux/mman.h>
+#  endif
+# endif
+#endif
+
```
madmajestro, line 45: should not be necessary, because all you need should already be included with sys/mman.h.

apc_mmap.c
```diff
@@ -51,7 +59,43 @@
...
+	// calculate log2 of huge page size
+	while (mmap_hugetlb_page_size) {
+		mmap_hugetlb_page_size >>= 1;
+		log2_page_size++;
+	}
+
+	if ((log2_page_size & MAP_HUGE_MASK) != log2_page_size) {
+		// maybe huge page size is too large
+		apc_warning("Invalid huge page size: %ld, using default huge page size", mmap_hugetlb_page_size);
```
madmajestro, line 86: mmap_hugetlb_page_size is 0 here.

apc_mmap.c
```diff
@@ -51,7 +59,43 @@
...
+		return MAP_HUGETLB;
+	}
+
+	return MAP_HUGETLB | ((unsigned int)log2_page_size << MAP_HUGE_SHIFT);
+# else
+	apc_warning("This system does not support HugeTLB pages");
+	return 0;
```
madmajestro, line 93: Can we check this earlier (early return) to reduce the length of the long `# if defined(MAP_HUGETLB) && defined(MAP_HUGE_MASK) && defined(MAP_HUGE_SHIFT)` block?

Imho it would also be nice to try to reduce / get rid most of the `#if defined(__linux__)` stuff. Perhaps it would be better to compile it all if APC_MMAP is set and let the user know that support for huge pages is not available if the system doesn't support it. What do you think?

madmajestro (commented):
This is a peer-review.

@nikic 
I don't know if you even want me to participate in the review. If not, please let me know, and I'll hold back a bit. Otherwise, I'm happy to help. The maintainer, of course, has the final say.

apc_mmap.c
KitanoTom: I deleted this part.

apc_mmap.c
KitanoTom: I fixed it.

apc_mmap.c
KitanoTom: I moved the checks for `MAP_HUGETLB`, `MAP_HUGE_MASK`, and `MAP_HUGE_SHIFT` to the top of the `apc_mmap_hugepage_flags` function.

I also removed the `#if defined(__linux__)` guard; the code now simply checks whether `APC_MMAP` is enabled. (and, the configuration parameter has been renamed from `apc_hugetlb_page_size` to `apc.mmap_hugepage_size`.)

apc_mmap.c
```diff
@@ -51,7 +51,35 @@
 # define MAP_ANON MAP_ANONYMOUS
 #endif
 
-void *apc_mmap(char *file_mask, size_t size)
+static int apc_mmap_hugepage_flags(zend_long hugepage_size)
+{
+#if !defined(MAP_HUGETLB) && !defined(MAP_HUGE_MASK) && !defined(MAP_HUGE_SHIFT)
```
madmajestro, line 56:
```suggestion
#if !defined(MAP_HUGETLB) || !defined(MAP_HUGE_MASK) || !defined(MAP_HUGE_SHIFT)
```

apc_mmap.c
```diff
@@ -51,7 +51,35 @@
...
+	apc_warning("This system does not support hugepages");
+	return 0;
+#else
+	zend_long page_size = hugepage_size;
+	int log2_page_size = -1;
+
+	if (hugepage_size == -1) return 0;           // not use hugepages
```
madmajestro, line 63: This check should be performed before `#if !defined(MAP_HUGETLB) && !defined(MAP_HUGE_MASK) && !defined(MAP_HUGE_SHIFT)` or on any system with mmap support but without HUGE_PAGES support, the warning "This system does not support hugepages" will occur.

KitanoTom:
I’ve updated the code based on your suggestion.

.github/workflows/config.yml
```diff
@@ -7,6 +7,8 @@ jobs:
           version: ["7.0", "7.1", "7.2", "7.3", "7.4", "8.0", "8.1", "8.2", "8.3", "8.4"]
     runs-on: ubuntu-latest
     steps:
+      - name: Set the hugepages parameter
+        run: sudo sh -c "echo 16 > /proc/sys/vm/nr_hugepages"
```
madmajestro, line 11: Lets reduce the number of HugePages in the pipeline to 1 (In apc_hugetlb.phpt the setting apc.shm_size must be changed to 2M).

madmajestro (commented):
During my testing, I noticed that the call to `munmap()` fails if `apc.shm_size` is not aligned to the HugePage size.

Example:
munmap() fails, if apc.shm_size=1M and apc.mmap_hugepage_size=2M

Therefore, we should check if it is aligned.

KitanoTom:
@madmajestro 
I’d like to get your thoughts on this.
To prevent the issue, we probably need a check like:
```
if (shm_size % hugepage_size) {
    zend_error_noreturn("apc.shm_size must be a multiple of apc.mmap_hugepage_size");
}
```

The problem is that when the `apc.mmap_hugepage_size` value is set to `default`, we have to determine the real hugepage size (e.g. by reading `/proc/meminfo`) which is a bit cumbersome.
Since this option’s purpose is to let users explicitly specify the hugepage size, I’m thinking about removing the `default` value altogether. What do you think?

KitanoTom:
I’ve removed the `none` and `default` options from `apc.mmap_hugepage_size` (set it to `0` to disable huge pages) and, also fixed the code so that throws an error if `apc.shm_size` is not an multiple of `apc.mmap_hugepage_size`.

madmajestro (commented):
We imho should not call `madvise()` if HugePages are explicitly configured.

That's the last thing I noticed.

madmajestro:
I just noticed that the test has not been added to `packets.xml` yet, so it does not run in the CI

KitanoTom:
@madmajestro 
I’ve made the changes according to your suggestions.
Thank you very much for your valuable assistance.

KitanoTom:
@nikic 
I've added some tests.

PR #561, opened 2025-05-15
Add origin reference to cache entries

madmajestro:
Cache entries now contain the address of the pointer/offset used to access them from a linked list. This can be either the address of the linked list start pointer or &entry->next of the previous entry. This is to prepare for defragmentation, as it allows entries to be moved in memory if the address containing the reference to the current entry is not available in the context.


madmajestro:
My plan is to write a defragmentation function in the SMA layer that iterates through all blocks for defragmentation. For each allocated block, a callback function from the cache layer is used to check whether a block can be moved and which changes the origin pointer to that entry. This callback function needs this origin reference to do the job.

If you have a better idea for the name of the field origin, please let me know...

madmajestro:
Superseded by #562

PR #562, opened 2025-05-23
Add defragmentation to apcu

madmajestro:
This adds defragmentation logic to apcu, which is performed during the default_expunge operation. It works by shifting all allocated blocks to the left (low addresses), allowing all free blocks to be coalesced to one larger free block on the right side.

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -160,29 +160,47 @@ static zend_bool apc_cache_entry_expired(
 		|| apc_cache_entry_soft_expired(cache, entry, t);
 }
 
-/* {{{ apc_cache_wlocked_remove_entry  */
-static void apc_cache_wlocked_remove_entry(apc_cache_t *cache, uintptr_t *entry_offset)
-{
-	apc_cache_entry_t *dead = ENTRYAT(*entry_offset);
+/* Inserts an entry into a linked list. The argument entry_offset must point either
+ * to entry->next of an existing entry or to the head pointer of a linked list. */
+static void apc_cache_wlocked_link_entry(apc_cache_t *cache, uintptr_t *entry_offset, apc_cache_entry_t *entry) {
+	entry->next = *entry_offset;
+	entry->prev = ENTRYOF(entry_offset);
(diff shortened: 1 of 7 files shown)


madmajestro:
A few words about the defragmentation's performance:

In my (rough) tests so far, it ran faster than expected, always taking less than 60% of the time required by the default expunge for the expiration loop. Furthermore, defragmentation had a positive effect on the insertion of new entries, as the subsequent search for free blocks worked faster.

Example:
I tested with 1 GB of shared memory and the smallest possible entry size with alternating TTL. In this case, the expiration loop took 0.345 seconds and the defragmentation took 0.193 seconds. These are typical values ​​I've observed in worst-case scenarios.
I used a Notebook with an AMD Ryzen 7 PRO 4750U with 32GB RAM for the test.

madmajestro:
> I think I need some help here. I can't figure out why the pipeline is failing. The pipeline works in my repository.

It seems that Appveyor had a temporary problem.

madmajestro:
Friendly ping.

apc_cache.c
```diff
@@ -160,29 +160,65 @@ static zend_bool apc_cache_entry_expired(
...
 
+/* apc_cache_wlocked_move_entry() is called during defragmentation, before an entry is moved to a new position. */
+static zend_bool apc_cache_wlocked_move_entry(apc_cache_t *cache, apc_cache_entry_t *old, apc_cache_entry_t *new) {
+	/* Check if the entry can be moved. */
+	if (old->ref_count > 0) {
+		return 0;
+	}
+
+	/* Change all references to this entry to the new position.
+	 * Since “next” is the 1st field of apc_cache_entry_t, the head pointer of the list
```
nikic, line 171: I don't really get why next being the 1st field matters here.

nikic: Thanks, I get it now.

nikic (approved):

madmajestro:
@nikic
Thanks a lot for your time!

PR #567, opened 2025-06-06
Fix Linux CI for PHP 7.x to not silently abort tests

### LINKED ISSUE #566: Pipelines do silently abort all PHP 7.x Linux tests
It seems that the Linux tests for all PHP 7.x versions have been silently skipped in the pipeline for some time now.

The Error `ERROR: Cannot run tests without CLI sapi.` is displayed and no test is executed. You can see this under `test` in every PHP 7 pipeline job.

I've traced this back to the point that the variable `PHP_EXECUTABLE` is set to `NONE` instead of `/usr/bin/php7.x` in the Makefile for all PHP 7 versions. But I haven't yet figured out the exact cause of this.

@nikic
Do you have any idea/hint for me how to fix the pipeline?

madmajestro:
All PHP 7.x Linux pipelines silently aborted the tests because the PHP executable could not be found. This was caused by the definition of "PHP_EXECUTABLE = NONE" in the Makefile, as "php-config --php-binary" returned "NONE" during ./configure. Since it seems time-consuming to fix the root cause, this is a temporary workaround until PHP 7.x is removed from the CI pipeline.

Closes #566

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -19,6 +19,8 @@ jobs:
         run: phpize
       - name: configure
         run: ./configure --enable-apcu-debug
+      - name: Fix missing PHP_EXECUTABLE in Makefile for PHP 7.x
+        run: sed -i -e 's/^PHP_EXECUTABLE = NONE$/PHP_EXECUTABLE = \/usr\/bin\/php/' Makefile
       - name: make
         run: make
       - name: test


madmajestro:
I've traced the cause back to that the version of the `php-config` script used by `shivammathur/setup-php` determines the available SAPIs differently in PHP 7.x and PHP 8.x.

While the available SAPIs are hardcoded in PHP 8.x, the PHP 7.x version determines them based on the installed Debian packages (which fails because the packages aren't installed).

Since fixing this seems time-consuming and PHP 7.x is already EOL, I propose a temporary workaround.

Issue #504, opened 2024-01-26
Introduce documentation for setting "apc.smart"


andrew-demb:
After dumping `apcu` settings with `php --ri apcu` I found a setting `apc.smart` (default = 0), which is not probably documented nor in `TECHNOTES.txt`, nor in php.net.

Please introduce documentation for this config entry.

PR #569, opened 2025-06-09
Update changelog

madmajestro:


joshtrichards:
> We should list the changes to apc.smart. I've added a bit more detail, as what apc.smart does is otherwise poorly documented.

Related / nearly closes out: #504

madmajestro:
Yes, but I suggest leaving #504 open until it is documented on php.net, which should be done with/after the next release.

PR #570, opened 2025-06-10
Add cache cleanup and defragmentation counts to apcu_cache_info()

madmajestro:
The cache cleanup and defragmentation counts are now available in the array returned by apcu_cache_info().

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -344,6 +344,8 @@ PHP_APCU_API apc_cache_t* apc_cache_create(apc_sma_t* sma, apc_serializer_t* ser
 	cache->header->nhits = 0;
 	cache->header->nmisses = 0;
 	cache->header->nentries = 0;
+	cache->header->ncleanups = 0;
+	cache->header->ndefragmentations = 0;
 	cache->header->nexpunges = 0;
 	cache->header->gc = 0;
 	cache->header->stime = time(NULL);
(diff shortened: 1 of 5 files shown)


TECHNOTES.txt
```diff
@@ -172,7 +172,8 @@ form of a quick-start guide to start hacking on APCu.
         zend_long nhits;                /* hit count */
         zend_long nmisses;              /* miss count */
         zend_long ninserts;             /* insert count */
-        zend_long nexpunges;            /* expunge count */
+        zend_long ncleanups;            /* default expunge count */
+        zend_long nexpunges;            /* real expunge count */
```
nikic, line 176: We should probably remove these copies of the struct definitions from this file and keep to a high level description.

madmajestro: Makes sense. Is it ok if i do this in a seperate PR?

apc_cache.c
```diff
@@ -1094,6 +1104,8 @@ PHP_APCU_API zend_bool apc_cache_info(zval *info, apc_cache_t *cache, zend_bool
 		add_assoc_double(info, "num_misses", (double) cache->header->nmisses);
 		add_assoc_double(info, "num_inserts", (double) cache->header->ninserts);
 		add_assoc_long(info,   "num_entries", cache->header->nentries);
+		add_assoc_long(info, "cleanups", cache->header->ncleanups);
+		add_assoc_long(info, "defragmentations", cache->header->ndefragmentations);
 		add_assoc_double(info, "expunges", (double) cache->header->nexpunges);
```
madmajestro, line 1109: I used `long` here and did not change the type of “expunges” for safety reasons. Is this a good choice, or is it better to use `double` for the new values ​​or change `expunges` to `long` to get consistent types?

nikic: I think changing the type of expunges to long would be fine.

madmajestro: Fixed.

PR #574, opened 2025-06-27
Update TECHNOTES

madmajestro:
A large portion of the TECHNOTES no longer reflected the current state of apcu. The TECHNOTES have therefore been revised to make it easier to get started. Since the copies of the struct definitions are not required to provide a general overview, they have been removed to avoid duplicate maintenance.


madmajestro:
As mentioned in #570, I initially just wanted to remove the struct definitions from the TECHNOTES. But after reading through the TECHNOTES, I decided to revise them, as many things were no longer correct.

Suggestions or better wording are welcome.

TECHNOTES.txt
```diff
@@ -3,326 +3,226 @@ APCu Quick-Start Braindump
...
+   perspective. PHP functions are all explained here: https://www.php.net/apcu
 
-2. Grab the current APC code from https://github.com/krakjoe/apcu
-    
-   apcu/php_apc.c has most of the code for the user-visible stuff.  It is
-   also a regular PHP extension in the sense that there are MINIT, MINFO, 
-   MSHUTDOWN, RSHUTDOWN, etc. functions.  
+2. Grab the current APCu code from https://github.com/krakjoe/apcu
+
+   Most of the user-visible stuff is primarily implemented apcu/php_apc.c.
```
nikic, line 11:
```suggestion
   Most of the user-visible stuff is implemented in apcu/php_apc.c.
```
(primarily redundant with most)

TECHNOTES.txt
```diff
@@ -3,326 +3,226 @@ APCu Quick-Start Braindump
...
+   It is also a regular PHP extension in the sense that there are MINIT, MINFO,
+   MSHUTDOWN, RSHUTDOWN, etc. functions.
 
 3. Build it.
 
    cd apcu
    phpize
-   ./configure --enable-apcu
+   ./configure --enable-apcu (or --enable-apcu-debug)
```
nikic, line 19: Is this "or"? I don't think --enable-apcu-debug implies --enable-acpu.

nikic (approved):

TECHNOTES.txt
```diff
@@ -3,326 +3,226 @@ APCu Quick-Start Braindump
...
-   - APCu can be configured to use third party serializers if they are compiled with support for apcu.
-     For example, `apc.serializer=igbinary` (https://github.com/igbinary/igbinary) can be used for generally faster unserialization and lower memory usage than apc.serializer=php 
-     (requires that igbinary be configured and compiled after APCu is installed)
+   This affects which serializer is used (Default: apc.serializer=php).
 
-If you made it to the end of this, you should have a pretty good idea of where things are in
-the code.  There is much more reading to do in headers ... good luck ...
+   APCu can be configured to use third-party serializers if they are compiled with support for apcu.
+   For example, `apc.serializer=igbinary` (https://github.com/igbinary/igbinary) can be used for
+   generally faster unserialization and lower memory usage than apc.serializer=php
```
madmajestro, line 214: Is that still correct? I simply copied the stuff about igbinary because my knowledge of the php and igbinary serializer isn't very deep.

nikic: I think so.

TECHNOTES.txt
madmajestro: Fixed. Yes, that makes sense. I was confused by the command `cp modules/apcu.so /usr/local/lib/php` in the technical notes a few lines above. I thought we should make sure we created a library so we could copy it.

But I just realized there are three ways to build apcu:
– Standalone / Out-of-Tree
– In-Tree, with phpize + configure + make in the `php-src/ext/apcu` directory (similar to Out-of-Tree)
– In-Tree, with buildconf + configure + make in the `php-src` directory

Your suggestion makes sense in all three cases:
The first and second cases create a shared library, while the third case creates a statically linked binary.

madmajestro:
The Windows pipeline seems to be a bit fragile...

PR #575, opened 2025-07-11
Fix relocatable representation of zend_empty_array

madmajestro:
When using the zend_empty_array optimization, the process-local address of zend_empty_array was stored in shared memory during persistence, which is inconsistent with the relocatable representation. To indicate the use of zend_empty_array during unpersist, the entry's starting address is now used during persist instead.


madmajestro:
I noticed this issue during my first test of APCu with PHP 8.5. I haven't noticed it in my previous tests because the zend_empty_array optimization apparently isn't used, since arrays always seem to be serialized. At least, that's the case in my PHP 8.3 environment. I haven't tested other PHP versions above 7.3 yet.

nikic:
@madmajestro I think it depends on the value of apc.serializer. We actually currently default to apc.serializer=php rather than default, so arrays always get serialized.

nikic (approved):

nikic:
Good point. It's unfortunate that we can't easily get an assertion-enabled PHP build in CI, unless we build it ourselves...

PR #576, opened 2025-07-26
Add reference and empty array to the defrag test

madmajestro:
A reference and an empty array have been added to the tested types. This allows the defragmentation test to verify whether references or the special case zend_empty_array work correctly after defragmentation.


madmajestro:
Right, I didn't take that into account. I removed the default serializer from the test and left the additional types.

PR #584, opened 2025-08-05
Fix race condition during allocation / defragmentation

madmajestro:
This fixes a race condition that can occur when a newly allocated entry is moved by defragmentation before it is inserted into the hash table.

### DIFF
--- a/apc_sma.h
+++ b/apc_sma.h
@@ -82,9 +82,12 @@ PHP_APCU_API void apc_sma_init(
 PHP_APCU_API void apc_sma_detach(apc_sma_t* sma);
 
 /*
-* apc_smap_api_malloc will allocate a block from the sma of the given size
+* apc_smap_api_malloc will allocate a block from the sma of the given size.
+* The init_callack() can be used to perform initializations that must be completed
+* before the lock of the sma layer is released.
 */
-PHP_APCU_API void* apc_sma_malloc(apc_sma_t* sma, size_t size);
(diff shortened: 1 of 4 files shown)


madmajestro:
@nikic 
Can you reproduce the hangs with this PR? And do you have a better idea/approach how to solve the problem?

apc_sma.h
```diff
@@ -82,9 +82,12 @@ PHP_APCU_API void apc_sma_init(
 PHP_APCU_API void apc_sma_detach(apc_sma_t* sma);
 
 /*
-* apc_smap_api_malloc will allocate a block from the sma of the given size
+* apc_smap_api_malloc will allocate a block from the sma of the given size.
+* The init_callack() can be used to perform initializations that must be completed
+* before the lock of the sma layer is released.
 */
-PHP_APCU_API void* apc_sma_malloc(apc_sma_t* sma, size_t size);
+typedef void (*apc_sma_malloc_init_f)(void *pointer); /* }}} */
```
nikic, line 89:
```suggestion
typedef void (*apc_sma_malloc_init_f)(void *pointer);
```

madmajestro:
Do you prepare a 5.1.26 release with #583 and #584?

madmajestro:
Thanks!

madmajestro:
@nikic 
What are these things (`/* }}} */`) actually needed for or by? Can I remove that in a future cleanup?

nikic:
They are a legacy code folding mechanism. It would be good to drop all of them.

PR #586, opened 2025-08-13
Remove all entries that cannot expire during default expunge

madmajestro:
When the cache is full, all entries that cannot expire are now removed by the first cleanup pass. This improves the performance of the default- expunge operation when apc.ttl = 0 and the application largely does not use an explicit TTL. Furthermore, this is closer to the behavior of APCu versions <= 5.1.24. However, unlike APCu <= 5.1.24, the first cleanup pass still retains all entries that expire via hard- or soft-ttl and have not yet expired, if apc.ttl = 0.

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -817,7 +817,8 @@ PHP_APCU_API zend_bool apc_cache_default_expunge(apc_cache_t* cache, size_t size
 		while (*entry_offset) {
 			apc_cache_entry_t *entry = ENTRYAT(*entry_offset);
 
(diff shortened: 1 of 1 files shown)


nikic:
I don't really understand the rationale for this behavior. Let's say I have a cache that is half filled with entries without TTL, and half filled with entries with expired TTL. Then currently, we'll clear out all the expired entries and leave the ones without TTL alone. Under the new implementation (if I understand correctly), we'd clear the whole cache -- why?

I think there are improvements to be made to prevent repeated whole cache scans, but I don't think this particular heuristic makes a lot of sense...

PR #587, opened 2025-08-13
Fix releasing of ref_count in apc_cache_store()

madmajestro:
In apc_cache_store(), the ref_count was decremented even when the entry needed to be deleted. This allowed the entry to be moved by defragmentation (by another process) before deletion. As a result, the pointer no longer pointed to the desired entry during deletion, causing segmentation faults.

### DIFF
--- a/apc_cache.c
+++ b/apc_cache.c
@@ -573,14 +573,16 @@ PHP_APCU_API zend_bool apc_cache_store(
 	php_apc_try {
 		ret = apc_cache_wlocked_insert(cache, entry, exclusive);
 	} php_apc_finally {
-		/* release entry, because the ref_count of a new entry is initialized to 1 during allocation */
-		apc_cache_entry_release(cache, entry);
 		apc_cache_wunlock(cache);
-	} php_apc_end_try();
 
-	if (!ret) {
-		free_entry(cache, entry);
-	}
+		if (ret) {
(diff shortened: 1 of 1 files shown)


apc_cache.c
```diff
@@ -573,14 +573,16 @@ PHP_APCU_API zend_bool apc_cache_store(
...
 
-	if (!ret) {
-		free_entry(cache, entry);
-	}
+		if (ret) {
+			/* release entry, because the ref_count of a new entry is initialized to 1 during allocation */
+			apc_cache_entry_release(cache, entry);
+		} else {
+			/* the entry mustn't be released before it is freed to prevent defragmentation from moving the entry */
+			free_entry(cache, entry);
```
madmajestro, line 583: Was there a reason why `free_entry()` was previously below `php_apc_end_try()`?

nikic (approved):

madmajestro:
@nikic
Do you plan to release version 5.1.27 today or tomorrow? As far as I know, a new PHP point release will probably be released tomorrow, and I'm afraid there might be problems again when the current version 5.1.26 is packaged and made available as a Debian package on packages.sury.org.

Also, #588 and #589 could significantly improve performance. I've tested both PRs for hours with various stress tests and haven't found any issues...

PR #589, opened 2025-08-19
Improve SMA allocation performance and lock contention

madmajestro:
The function sma_allocate() now inserts the unused part of a split block at the beginning of the free list, instead of reinserting it at the position of the original block. This prevents large blocks from falling behind in the list when many small blocks are inserted at the beginning of the list by sma_deallocate(). Benchmarks show that this improves allocation performance and reduces the time the SMA lock must be held.

### DIFF
--- a/apc_sma.c
+++ b/apc_sma.c
@@ -98,6 +98,21 @@ struct block_t {
 /* How many extra blocks to check for a better fit */
 #define BEST_FIT_LIMIT 3
 
+static inline void link_free_block_at_start(sma_header_t *smaheader, block_t *cur) {
+	block_t *dst = BLOCKAT(ALIGNWORD(sizeof(sma_header_t)));
+
+	/* insert cur as first block in the free list */
+	cur->fnext = dst->fnext;
+	cur->fprev = OFFSET(dst);
+	dst->fnext = OFFSET(cur);
+	BLOCKAT(cur->fnext)->fprev = dst->fnext;
+}
+
+static inline void unlink_free_block(sma_header_t *smaheader, block_t *cur) {
+	BLOCKAT(cur->fprev)->fnext = cur->fnext;
+	BLOCKAT(cur->fnext)->fprev = cur->fprev;
+}
+
 static inline block_t *find_block(sma_header_t *smaheader, size_t realsize) {
 	block_t *cur = BLOCKAT(ALIGNWORD(sizeof(sma_header_t)));
 	block_t *found = NULL;
@@ -150,8 +165,7 @@ static APC_HOTSPOT size_t sma_allocate(sma_header_t *smaheader, size_t size)
 	}
 
 	/* remove cur from the list of free blocks */
-	BLOCKAT(cur->fprev)->fnext = cur->fnext;
-	BLOCKAT(cur->fnext)->fprev = cur->fprev;
+	unlink_free_block(smaheader, cur);
 
 	if (cur->size >= realsize && cur->size < (realsize + smaheader->min_block_size)) {
 		/* cur is big enough for realsize, but too small to split */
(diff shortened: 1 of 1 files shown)


madmajestro:
@nikic
While working on a size-binning allocator, I noticed that we can significantly increase the performance of write operations in APCu by changing the SMA so that it tries to keep a large block at the beginning of the free list. The results of this small change are impressive. Furthermore, I haven't found any downsides in my tests. My concern here was increased fragmentation, but I didn't detect that. The number of cleanups and defragmentations was only higher because more entries were added within a shorter period of time. I checked this by adding an usleep() to the test loop. The test script used and some test results will follow...

madmajestro:
```
... continuation ...

php -d apc.enable_cli -d apc.shm_size=512M stress_test.php

APCu (current master-branch):

Operations: 16000000, Seconds: 1107.05, Ops/Sec: 14452.79
Cleanups: 0, Defragmentations: 0, Expunges: 0

APCu including the change:

Operations: 16000000, Seconds: 76.68, Ops/Sec: 208667.94
Cleanups: 0, Defragmentations: 0, Expunges: 0

Conclusion: The new version performs 14,44 times better.

-----

php -d apc.enable_cli -d apc.shm_size=1024M stress_test.php

APCu (current master-branch):

CANCELED
I canceled the test when I saw the very first message with 100,000 operations after exactly 21 minutes (because my time was limited). We can therefore roughly assume that the test would probably have continued running for more than 10 times as long.

APCu including the change:

Operations: 16000000, Seconds: 58.78, Ops/Sec: 272209.01
Cleanups: 0, Defragmentations: 0, Expunges: 0

Conclusion: The new version performs (estimated!) more than 200 times better (21 * 60 * 10 / 58.78 = 214,36).
```

### Noticeable findings during the first tests:

- It's noticeable that the newer version becomes faster as the memory size increases. This is because apc.entries_hint automatically scales up with the memory. I have verified this by running the test with `php -d apc.enable_cli -d apc.shm_size=256M -d apc.entries_hint=512000 stress_test.php`, which only took 31.73 seconds.
- The master version, on the other hand, becomes increasingly slower because the lock contention of the SMA lock slows everything down as the list of free blocks grows.


### Another test run with the following modifications:
- This test was run with `$keyLen = random_int(1, 100);` (to decrease collisions)
- This test was run without explicit TTL

```
php -d apc.enable_cli -d apc.shm_size=32M stress_test.php

APCu (current master-branch):

Operations: 16000000, Seconds: 21.05, Ops/Sec: 760124.85
Cleanups: 702, Defragmentations: 387, Expunges: 350

APCu including the change:

Operations: 16000000, Seconds: 21.04, Ops/Sec: 760293.26
Cleanups: 705, Defragmentations: 403, Expunges: 350

Conclusion: The performance is nearly the same. Sometimes one version wins, sometimes the other.

-----

php -d apc.enable_cli -d apc.shm_size=1024M stress_test.php

APCu (current master-branch):

Operations: 16000000, Seconds: 202.86, Ops/Sec: 78873.58
Cleanups: 20, Defragmentations: 10, Expunges: 10

APCu including the change:

Operations: 16000000, Seconds: 30.31, Ops/Sec: 527818.82
Cleanups: 21, Defragmentations: 12, Expunges: 10

Conclusion: The new version performs 6.69 times better.
```

### Another test run with the following modifications:
- This test was run with `$valLen = random_int(1, 4000);` (to use more different allocation sizes)
- This test was run with `$keyLen = random_int(1, 100);` (to decrease collisions)
- This test was run without explicit TTL

```
php -d apc.enable_cli -d apc.shm_size=32M stress_test.php

APCu (current master-branch):

Operations: 16000000, Seconds: 24.39, Ops/Sec: 656051.94
Cleanups: 1109, Defragmentations: 88, Expunges: 1068

APCu including the change:

Operations: 16000000, Seconds: 23.68, Ops/Sec: 675535.64
Cleanups: 1115, Defragmentations: 82, Expunges: 1067

Conclusion: The performance is similar. But most times the new version wins.

-----

php -d apc.enable_cli -d apc.shm_size=1024M stress_test.php

APCu (current master-branch):

Operations: 16000000, Seconds: 78.6, Ops/Sec: 203554.88
Cleanups: 66, Defragmentations: 33, Expunges: 33

APCu including the change:

Operations: 16000000, Seconds: 37.16, Ops/Sec: 430564.24
Cleanups: 66, Defragmentations: 34, Expunges: 33

Conclusion: The new version performs 2.12 times better.
```

Based on the test results, I suspect that with this change, APCu can be safely used with significantly more memory without fear of a complete drop in performance. Furthermore, this version appears to offer better write performance in several scenarios.

madmajestro:
Friendly ping

Issue #582, opened 2025-08-04
apcu_fetch() hangs on PHP 8.1 and Debian 11


klausi:
After updating to apcu 5.1.25 via the Sury sources https://deb.sury.org/ with PHP 8.1 on Debian 11 our busy sites went down.

Int PHP FPM slow log we can see apcu_fetch() hanging in our Drupal 10 site
```
[04-Aug-2025 10:52:29]  [pool www] pid 6804
script_filename = /var/www/example/site/backend/web/index.php
[0x00007f8b90815f40] apcu_fetch() /var/www/example/site/backend/vendor/composer/ClassLoader.php:452
[0x00007f8b90815ea0] findFile() /var/www/example/site/backend/vendor/composer/ClassLoader.php:425
[0x00007f8b90815e10] loadClass() /var/www/example/site/backend/web/core/lib/Drupal/Component/DependencyInjection/Container.php:261
[0x00007f8b90815cd0] createService() /var/www/example/site/backend/web/core/lib/Drupal/Component/DependencyInjection/Container.php:453
[0x00007f8b90815be0] resolveServicesAndParameters() /var/www/example/site/backend/web/core/lib/Drupal/Component/DependencyInjection/Container.php:275
[0x00007f8b90815aa0] createService() /var/www/example/site/backend/web/core/lib/Drupal/Component/DependencyInjection/Container.php:179
[0x00007f8b908159d0] get() /var/www/example/site/backend/web/core/lib/Drupal/Component/DependencyInjection/Container.php:438
[0x00007f8b908158e0] resolveServicesAndParameters() /var/www/example/site/backend/web/core/lib/Drupal/Component/DependencyInjection/Container.php:239
[0x00007f8b908157a0] createService() /var/www/example/site/backend/web/core/lib/Drupal/Component/DependencyInjection/Container.php:179
[0x00007f8b908156d0] get() /var/www/example/site/backend/web/core/lib/Drupal/Component/DependencyInjection/Container.php:438
[0x00007f8b908155e0] resolveServicesAndParameters() /var/www/example/site/backend/web/core/lib/Drupal/Component/DependencyInjection/Container.php:239
[0x00007f8b908154a0] createService() /var/www/example/site/backend/web/core/lib/Drupal/Component/DependencyInjection/Container.php:179
[0x00007f8b908153d0] get() /var/www/example/site/backend/web/core/lib/Drupal/Component/DependencyInjection/Container.php:438
[...]

PR #592, opened 2025-08-28
Prepare apcu 5.1.27

madmajestro:


madmajestro:
I hope this helps closing #582. I chose today's date because a new PHP point release is being released today, and many are updating their systems. If you don't like a new version being prepared by a contributor, please let me know and I won't do this again in the future.

PR #593, opened 2025-09-12
Remove test bug63224.phpt

madmajestro:
The test has been removed because it serves no purpose anymore, as the opcode caching functionality was removed from the project long ago. Furthermore, the magic method __sleep() is deprecated in PHP 8.5, causing this test to fail the CI pipeline starting with PHP 8.5.


madmajestro:
Friendly ping

nikic (commented):
I think this test can be deleted entirely: It does not actually use apcu at all. This seems like a leftover from back when apc was also doing opcode caching.

madmajestro:
Makes sense. Done.

nikic (approved):

nikic:
Looking at the test directory, we might be able to delete `tests/get_included_files_inc*.inc` and `tests/*/abc.data` as well.

madmajestro:
I will provide a PR which removes the `tests/get_included_files_inc*.inc` stuff, but  it looks like the `abc.data` files are still used by some tests which test preloading:
- bad/abc.data: ghbug176.phpt
- data/abc.data: apc54_014.phpt

madmajestro:
see #594

PR #601, opened 2025-11-15
apc.php / gzdecode rss

SailorMax:
decode rss content if it compressed

### DIFF
--- a/apc.php
+++ b/apc.php
@@ -1117,6 +1117,10 @@ EOB;
     } else {
         $apcversion = phpversion('apcu');
 
+        if ($rss[0] !== '<') {
+            $rss = gzdecode($rss);
+        }
+
         preg_match('!<title>APCu ([0-9.]+)</title>!', $rss, $match);
         echo '<tr class="tr-0 center"><td>';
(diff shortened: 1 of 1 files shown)


apc.php
```diff
@@ -1117,6 +1117,10 @@ function block_sort($array1, $array2)
     } else {
         $apcversion = phpversion('apcu');
 
+        if ($rss[0] !== '<') {
+            $rss = gzdecode($rss);
+        }
```
madmajestro:
```suggestion
        if (ord($rss[0]) === 0x1f && ord($rss[1]) === 0x8b) {
            $rss = gzdecode($rss);
        }
```

- I propose to check for the magic numbers 0x1f and 0x8b to identify the data as a gzip stream. Imho, this is more precise and it allows us to pick different decompression algorithms if necessary someday.
- A decompression failure (if gzdecode() returns false) should be handled properly. Perhaps it makes sense to do this a few lines earlier (before line 1115)?

nikic: Agree that it's better to check the magic numbers. (Even better would probably to check for the content-encoding response header.)

Issue #615, opened 2026-04-09
PHP 8.6 failed to build


andypost:
After  https://github.com/php/php-src/commit/04f06e8cf28e1a9dff54b84fb10bc403bb2f91be APCu fails to build

As it supports 7.2+ I bet it safe to replace `EMPTY_SWITCH_DEFAULT_CASE` with  `default: ZEND_UNREACHABLE() ` directly

ref https://github.com/php/php-src/pull/20063

log
```
/tmp/pear/temp/apcu/apc_persist.c: In function 'apc_persist_calc_zval':
/tmp/pear/temp/apcu/apc_persist.c:247:17: error: implicit declaration of function 'EMPTY_SWITCH_DEFAULT_CASE' [-Wimplicit-function-declaration]
  247 |                 EMPTY_SWITCH_DEFAULT_CASE()
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/pear/temp/apcu/apc_persist.c:247:44: error: expected ';' before '}' token
  247 |                 EMPTY_SWITCH_DEFAULT_CASE()
      |                                            ^
      |                                            ;
  248 |         }
      |         ~
/tmp/pear/temp/apcu/apc_persist.c: In function 'apc_persist_copy_zval_impl':
/tmp/pear/temp/apcu/apc_persist.c:453:44: error: expected ';' before '}' token
  453 |                 EMPTY_SWITCH_DEFAULT_CASE()
      |                                            ^
      |                                            ;
  454 |         }
      |         ~
/tmp/pear/temp/apcu/apc_persist.c: In function 'apc_persist_calc_zval':
/tmp/pear/temp/apcu/apc_persist.c:249:1: warning: control reaches end of non-void function [-Wreturn-type]
  249 | }
      | ^
make: *** [Makefile:245: apc_persist.lo] Error 1
```

nikic:
Fixed by https://github.com/krakjoe/apcu/pull/616.

PR #616, opened 2026-04-09
Fix build against PHP 8.6 (EMPTY_SWITCH_DEFAULT_CASE removed)

andypost:
Fix for #615

### DIFF
--- a/apc.h
+++ b/apc.h
@@ -67,6 +67,16 @@
 #include "php.h"
 #include "main/php_streams.h"
 
+/* ZEND_UNREACHABLE() was added in PHP 8.0. Provide fallback for PHP 7.x.
+ * Remove this block when APCu drops support for PHP < 8.0. */
+#if PHP_VERSION_ID < 80000
+# if ZEND_DEBUG
+#  define ZEND_UNREACHABLE() do {ZEND_ASSERT(0); ZEND_ASSUME(0);} while (0)
+# else
+#  define ZEND_UNREACHABLE() ZEND_ASSUME(0)
+# endif
+#endif
+
 /* console display functions */
 PHP_APCU_API void apc_error(const char *format, ...) ZEND_ATTRIBUTE_FORMAT(printf, 1, 2);
 PHP_APCU_API void apc_warning(const char *format, ...) ZEND_ATTRIBUTE_FORMAT(printf, 1, 2);
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -11,7 +11,7 @@ jobs:
   ubuntu:
     strategy:
       matrix:
-          version: ["7.0", "7.1", "7.2", "7.3", "7.4", "8.0", "8.1", "8.2", "8.3", "8.4", "8.5"]
+          version: ["7.0", "7.1", "7.2", "7.3", "7.4", "8.0", "8.1", "8.2", "8.3", "8.4", "8.5", "8.6"]
     runs-on: ubuntu-latest
     steps:
       - name: Set the hugepages parameter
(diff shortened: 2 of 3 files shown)


andypost:
yes, it was 8.0 https://github.com/php/php-src/commit/92c4b0651368a4135f692af48f4716c96de4bd92

madmajestro (commented):
It's probably better to replace the two occurrences  of `EMPTY_SWITCH_DEFAULT_CASE()` with `default: ZEND_UNREACHABLE();` as it was done in https://github.com/php/php-src/commit/04f06e8cf28e1a9dff54b84fb10bc403bb2f91be

madmajestro (changes requested):
Maybe i should explain my last comment a little bit more:
- Your code doesn't fix the problem for PHP 8.6 because `php_apc.h` is not included by `apc_persist.c` where `EMPTY_SWITCH_DEFAULT_CASE` is used.
- It would imho be better if we can get rid of `EMPTY_SWITCH_DEFAULT_CASE` to be able to write new code as done in php-src.
- Therefore it would imho be better to insert compatibility code which defines ZEND_UNREACHABLE for PHP < 8.0 in `apc.h`. This has the useful side effect that, due to the check for `PHP_VERSION_ID`, the code can be easily found and removed if someday  support for PHP < 8 is dropped.

andypost:
Replaced commit with fallback for 7.x in `apc.h` and replaced usage as internals suggest

nikic:
Can you please also add 8.6 to the CI matrix?

andypost:
Added 8.6 but it failed for windows to fetch https://github.com/krakjoe/apcu/actions/runs/24210376612/job/70677213530

maybe needs manual restart

andypost:
Transition to `php-windows-builder/extension@v1` helped, all green

apc.h
```diff
@@ -67,6 +67,14 @@
 #include "php.h"
 #include "main/php_streams.h"
 
+/* ZEND_UNREACHABLE() was added in PHP 8.0. Provide fallback for PHP 7.x.
+ * Remove this block when APCu drops support for PHP < 8.0. */
+#if PHP_VERSION_ID < 80000
+# ifndef ZEND_UNREACHABLE
+#  define ZEND_UNREACHABLE() ZEND_ASSUME(0)
+# endif
```
madmajestro:
```suggestion
# if ZEND_DEBUG
#  define ZEND_UNREACHABLE() do {ZEND_ASSERT(0); ZEND_ASSUME(0);} while (0)
# else
#  define ZEND_UNREACHABLE() ZEND_ASSUME(0)
# endif
```

- I propose to keep the assert in debug mode as php-src does.
- The `#ifndef ZEND_UNREACHABLE` should not be necessary, as PHP < 8.0 doesn't implement `ZEND_UNREACHABLE` and it's probably better to get informed otherwise.

madmajestro (commented):
I like the approach of using `php/php-windows-builder`. However, I've noticed that the names and internal structure of the build artifacts have changed. I'm not sure if that's important.

andypost:
artifacts are handled now `php/php-windows-builder` and I bet tDLL/PDB are still included in the artifacts, just under the builder's naming scheme which supposed to be common for PECL

> `apcu-8.5-x64-nts` to `php_apcu-616-8.5-nts-vs17-x86_64`

madmajestro (approved):
From my perspective:
- The C-Code changes are in shape now.
- The switch to `php/php-windows-builder` is maybe somewhat out of scope of this PR, but i like the approach and in the long term it makes sense. So if @nikic is fine with it we should just do it.

-> LGTM

Thank you!

nikic (approved):

PR #617, opened 2026-04-10
apc mmap: fix format specifier for hugepage_size in error message

arshidkv12:


apc_mmap.c
```diff
@@ -118,7 +118,7 @@ void *apc_mmap(char *file_mask, size_t size, zend_long hugepage_size)
 
 	if ((long)shmaddr == -1) {
 		if (hugepage_size) {
-			zend_error_noreturn(E_CORE_ERROR, "apc_mmap: Failed to mmap %zu bytes with hugepage size %ld. apc.shm_size may be too large, apc.mmap_hugepage_size may be invalid, or the system lacks sufficient reserved hugepages.", size, hugepage_size);
+			zend_error_noreturn(E_CORE_ERROR, "apc_mmap: Failed to mmap %zu bytes with hugepage size %lld. apc.shm_size may be too large, apc.mmap_hugepage_size may be invalid, or the system lacks sufficient reserved hugepages.", size, hugepage_size);
```
madmajestro, line 121:
```suggestion
			zend_error_noreturn(E_CORE_ERROR, "apc_mmap: Failed to mmap %zu bytes with hugepage size " ZEND_LONG_FMT ". apc.shm_size may be too large, apc.mmap_hugepage_size may be invalid, or the system lacks sufficient reserved hugepages.", size, hugepage_size);
```

This causes compiler warnings on my 64 bit linux system and it should cause problems when compiling on 32 bit systems, too. What system do you use? Is it Windows 64-bit?

The preferred specifier for `zend_long` is `ZEND_LONG_FMT`, which should be portable.

madmajestro (changes requested):

apc_mmap.c
arshidkv12: Mac OS 64-bit. I checked with clang-tidy.

madmajestro: Does clang-tidy complain when using ZEND_LONG_FMT?

I think it's not worth fixing only one occurrence with a PR . Do you like to fix all occurrences? Otherwise, I'd be happy to take over and create a pull request that addresses all occurrences.

madmajestro:
@arshidkv12
Should be fixed by #618. Please check if the problem still exists.

PR #618, opened 2026-04-10
Use proper format specifiers

madmajestro:
ZEND_LONG_FMT should be used for the type zend_long to avoid problems on platforms where long and zend_long have a different size. In addition, %zu is now consistently used to format size_t.


madmajestro:
This supersedes #617

nikic (approved):

PR #619, opened 2026-04-11
Update to actions/checkout@v6

madmajestro:
GitHub action checkout@v4 uses Node.js 20, which is deprecated and will be removed from the runner image later this year. It is replaced by checkout@v6, which uses Node.js 24.


madmajestro:
The pipeline runs recently contained annotations indicating the problem.

PR #621, opened 2026-04-30
Replace XtOffsetOf() with offsetof() for PHP 8.6

andypost:
PHP 8.6 removed `XtOffsetOf()`, which was an alias for the standard C `offsetof()` macro. Replaced all usages in `apc_cache.h` and `apc_iterator`  with` offsetof()` directly.

Ref https://github.com/php/php-src/commit/7114314c5a96c362b95663f7e7c9184586721f58

>  The XtOffsetOf() alias of C’s offsetof() macro has been removed. Use  offsetof() directly.

### DIFF
--- a/apc_cache.h
+++ b/apc_cache.h
@@ -295,7 +295,7 @@ static inline void apc_cache_runlock(apc_cache_t *cache) {
 }
 
 /* APC_ENTRY_SIZE takes into account the trailing key-string + terminating 0-byte */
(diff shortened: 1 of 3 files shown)


madmajestro (approved):
LGTM

nit:
The commit title and/or body could contain the change as well as the reason in plain text.

andypost:
Thanks, I was fixing drupal's nightly jobs and hotfixed but there's only this change so I will rename PR and amend commit

madmajestro:
@andypost

Thank you!

nikic (approved):

PR #623, opened 2026-05-10
Add new release workflow to publish files required by PIE

### LINKED ISSUE #568: about apcu-8.4-x64-nts.zip and apcu-8.4-x64-ts.zip
Why don't you publish the new apcu-8.4-x64-nts.zip and apcu-8.4-x64-ts.zip inside the 5.1.24 (Assets) release?

https://pecl.php.net/package/APCu/5.1.24/windows

thk

madmajestro:
This commit introduces a new release workflow which is triggered if a new release or prerelease is published on GitHub. It builds zip-files with prebuilt Windows Dlls and uploads them as artifacts to the release. These artifacts are required by PIE (PHP Installer for Extensions) to function correctly for Windows users. The CI pipeline has been modified to build and test the extension for all Windows PHP versions for which installation packages are available via PECL or PIE.

Closes #568

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -1,5 +1,10 @@
 name: Build and Test
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    # Trigger on all branches, but not on tags, to prevent CI runs if a release is published (which pushes a tag)
+    branches: ['**']
+
 jobs:
   common:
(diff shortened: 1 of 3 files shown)


madmajestro:
This also introduces additional builds for older PHP versions on Windows, but I think that as long as we support them - and they are available on PECL) - we should test them in the CI as well.

PR #624, opened 2026-06-01
Avoid unnecessary hash value computation for serialized data

madmajestro:
For entries storing the data in a serialized form, the hash value of the serialized data string is no longer computed, as the hash is not needed for the unserialization. Synthetic benchmarks show a significant performance gain when saving larger objects or arrays by using the PHP serializer (which is the default).

### DIFF
--- a/apc_persist.c
+++ b/apc_persist.c
@@ -294,19 +294,18 @@ static zend_string *apc_persist_copy_cstr(
 	ZSTR_LEN(str) = buf_len;
 	memcpy(ZSTR_VAL(str), orig_buf, buf_len);
 	ZSTR_VAL(str)[buf_len] = '\0';
-	zend_string_hash_val(str);
 
 	return str;
 }
 
 static zend_string *apc_persist_copy_zstr_no_add(
-		apc_persist_context_t *ctxt, const zend_string *orig_str) {
+		apc_persist_context_t *ctxt, zend_string *orig_str) {
 	return apc_persist_copy_cstr(
-		ctxt, ZSTR_VAL(orig_str), ZSTR_LEN(orig_str), ZSTR_H(orig_str));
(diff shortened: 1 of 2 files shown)


madmajestro:
I roughly measured the performance difference by using the following test script:

```
<?php

function perf_test(int $iterations, int $dataLen) {
    $data = [0 => random_bytes($dataLen)];

    apcu_delete('1');

    $start_time = microtime(true);

    for ($i = 0; $i < $iterations; $i++) {
        apcu_store('1', $data);
    }

    $elapsed_time = microtime(true) - $start_time;

    echo "Iterations: " . $iterations . ", Length: " . $dataLen . ", Seconds: " . round($elapsed_time, 2) . "\n";
}

perf_test(1000000, 10);
perf_test(1000000, 100);
perf_test(1000000, 1000);
perf_test(1000000, 10000);
```

Test-Environment:
Debian 13 (Trixie) with PHP 8.5.6, AMD Ryzen 7 PRO 4750U

Test results without the change (master, 07be608c7fa4a2ad48b1f1a430d3b63f7817483b):
Iterations: 1000000, Length: 10, Seconds: 0.65
Iterations: 1000000, Length: 100, Seconds: 0.66
Iterations: 1000000, Length: 1000, Seconds: 0.98
Iterations: 1000000, Length: 10000, Seconds: 3.76

Test results (master + including the change):
Iterations: 1000000, Length: 10, Seconds: 0.61
Iterations: 1000000, Length: 100, Seconds: 0.61
Iterations: 1000000, Length: 1000, Seconds: 0.67 <-- ~30% faster
Iterations: 1000000, Length: 10000, Seconds: 1.06 <-- ~3.5 times faster

Conclusion:
The performance improvement is significant (especially for larger serialized buffers).

madmajestro:
`apc_persist_copy_cstr()` is called by `apc_persist_copy_serialize()` with `0` as hash, which caused the hash to be computed in `apc_persist_copy_cstr()` prior to this change.

PR #627, opened 2026-06-30
Don't unserialize iterator values while holding the cache lock

iliaal:
APCUIterator builds each item, including unserializing the stored value, while holding the cache read lock. Objects are always stored serialized, so this runs `__wakeup()` / `Serializable::unserialize()` under the lock. If that userland code re-enters apcu with a write operation, it takes the write lock while the read lock is held and the worker self-deadlocks. On recursive-mutex builds the nested write succeeds instead and can free or relocate the entry mid-iteration, a use-after-free.

Reproduce (hangs the worker):

```php
class C { function __wakeup() { apcu_store('x', 1); } }
apcu_store('obj', new C());
foreach (new APCUIterator(null, APC_ITER_VALUE) as $_) {}
```

`apc_cache_fetch()` already avoids this: it increfs the entry, releases the lock, then unserializes. This applies the same pattern to the iterator. Under the lock it pins (increfs) the matching entries, releases the lock, then builds the items and drops the references. Pinning blocks defragmentation and freeing (`apc_cache_wlocked_move_entry` refuses `ref_count > 0`), so the entry pointers stay valid across the unlock. Includes a regression test.

### DIFF
--- a/apc_iterator.c
+++ b/apc_iterator.c
@@ -93,7 +93,9 @@ static apc_iterator_item_t* apc_iterator_item_ctor(
 		zend_hash_add_new(ht, apc_str_access_time, &zv);
 	}
 	if (APC_ITER_REFCOUNT & iterator->format) {
-		ZVAL_LONG(&zv, entry->ref_count);
+		/* The iterator holds one reference to pin the entry during construction;
+		 * exclude it so the reported count reflects the cache-visible value. */
+		ZVAL_LONG(&zv, entry->ref_count - 1);
 		zend_hash_add_new(ht, apc_str_ref_count, &zv);
 	}
 	if (APC_ITER_MEM_SIZE & iterator->format) {
@@ -127,6 +129,7 @@ static void apc_iterator_free(zend_object *object) {
 	}
 
 	apc_stack_destroy(iterator->stack);
+	apc_stack_destroy(iterator->entries);
 
 	if (iterator->regex) {
 		zend_string_release(iterator->regex);
@@ -153,6 +156,7 @@ zend_object* apc_iterator_create(zend_class_entry *ce) {
 
 	iterator->initialized = 0;
 	iterator->stack = NULL;
+	iterator->entries = NULL;
 	iterator->regex = NULL;
 	iterator->search_hash = NULL;
 	iterator->obj.handlers = &apc_iterator_object_handlers;
@@ -199,7 +203,7 @@ static int apc_iterator_check_expiry(apc_cache_t* cache, apc_cache_entry_t *entr
 static size_t apc_iterator_fetch_active(apc_iterator_t *iterator) {
 	apc_cache_t *cache = apc_user_cache;
 	size_t count = 0;
-	apc_iterator_item_t *item;
+	int i;
 	time_t t = apc_time();
 
 	while (apc_stack_size(iterator->stack) > 0) {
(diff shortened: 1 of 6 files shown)


madmajestro:
Thanks for your contribution! At first glance, this looks like a problem worth taking a closer look at.

I'm currently on vacation for two weeks. I'll probably have not much time for reviews during this period, but I'll try.

tests/iterator_012.phpt
madmajestro, line 1: - This file must be added to `package.xml`
- The file could be named `iterator_012.phpt` to align it's name with the other iterator tests.

apc_iterator.c
```diff
@@ -210,27 +214,49 @@ static size_t apc_iterator_fetch_active(apc_iterator_t *iterator) {
...
+		php_apc_try {
+			while (count <= iterator->chunk_size && iterator->slot_idx < cache->nslots) {
+				uintptr_t entry_offset = cache->slots[iterator->slot_idx];
+				while (entry_offset) {
+					apc_cache_entry_t *entry = ENTRYAT(entry_offset);
+					if (apc_iterator_check_expiry(cache, entry, t)) {
+						if (apc_iterator_search_match(iterator, entry)) {
+							count++;
+							apc_cache_entry_incref(cache, entry);
+							apc_stack_push(entries, entry);
```
madmajestro: `apc_stack_push()` can call `erealloc()` and if my understanding is right this can cause a bailout in oom situations. Therefore, `apc_stack_push()` should be called before `apc_cache_entry_incref()`. Otherwise the incremented reference counter of the entry which causes the bailout will not be decremented in the finally block.

apc_iterator.c
```diff
@@ -210,27 +214,49 @@ static size_t apc_iterator_fetch_active(apc_iterator_t *iterator) {
 		return count;
 	}
 
+	entries = apc_stack_create(iterator->chunk_size);
```
madmajestro: Creating a new stack (2 x emalloc()) while holding the lock seems suboptimal. It should be created either before acquiring the lock or when the iterator is created (iterator contains two stacks). The latter is probably better, because it avoids the need to create / destroy a new stack for each iterator fetch / chunk. 

Looking at this i wonder if we need two stacks at all:
- It should be possible to use one stack for both. All we need is to implement a new function `apc_stack_set()` in `apc_stack.[ch]` which enables us to set (replace) the value on the stack.
- The loop which uses `apc_iterator_item_ctor()` to construct the items could decrement the reference counter and replace the pointer to the entry by a pointer to an item.
- it must be ensured that no reference to an entry remains on the stack.

I would like to keep the performance impact of this change as low as possible. TBH it has the chance to improve the situation in some cases due to reduced lock contention.

madmajestro (changes requested):

apc_iterator.c
```diff
@@ -210,27 +214,49 @@ static size_t apc_iterator_fetch_active(apc_iterator_t *iterator) {
...
+			}
+		} php_apc_finally {
+			iterator->stack_idx = 0;
+			apc_cache_runlock(cache);
+		} php_apc_end_try();
+
+		/* Build the items with the lock released. */
+		for (i = 0; i < apc_stack_size(entries); i++) {
+			item = apc_iterator_item_ctor(iterator, apc_stack_get(entries, i));
+			if (item) {
```
madmajestro: This if should be removed, because item should always be present. A bailout should occur in apc_iterator_item_ctor() otherwise.

tests/iterator_012.phpt
iliaal: Renamed to `iterator_012.phpt` and added it to `package.xml`.

apc_iterator.c
iliaal: Right, with incref first a bailout in the push would leak the pin. Reordered so `apc_stack_push()` runs before the incref.

apc_iterator.c
iliaal: Removed. `apc_iterator_item_ctor()` bails on OOM rather than returning NULL, so the check was dead.

apc_iterator.c
iliaal: Moved the entries stack onto the iterator, created once at construction and cleared (not reallocated) per fetch, so there's no allocation under the lock and no per-chunk create/destroy.

I'd keep it separate from `iterator->stack` rather than folding both into one via `apc_stack_set()`: the two have different bailout lifetimes. Entries must be fully released on any bailout; items get handed to the caller. With one stack, a bailout inside `apc_iterator_item_ctor()` mid-build leaves the `finally` unable to tell which slots are still entries to release versus finished items. Two stacks encode that boundary structurally.

madmajestro (commented):
The only change I can see is that the test has been renamed. Did you perhaps not commit or push all changes?

iliaal:
> The only change I can see is that the test has been renamed. Did you perhaps not commit or push all changes?

Helps to do proper push, the changes are there now, sorry 🤦

apc_iterator.c
```diff
@@ -210,27 +215,44 @@ static size_t apc_iterator_fetch_active(apc_iterator_t *iterator) {
...
+		for (i = 0; i < apc_stack_size(entries); i++) {
+			apc_stack_push(iterator->stack, apc_iterator_item_ctor(iterator, apc_stack_get(entries, i)));
 		}
 	} php_apc_finally {
-		iterator->stack_idx = 0;
-		apc_cache_runlock(cache);
+		for (i = 0; i < apc_stack_size(entries); i++) {
+			apc_cache_entry_release(cache, apc_stack_get(entries, i));
+		}
+		apc_stack_clear(entries);
```
madmajestro:
```suggestion
		while (apc_stack_size(entries) > 0) {
			apc_cache_entry_release(cache, apc_stack_pop(entries));
		}
```

apc_iterator.c
```diff
@@ -329,6 +363,7 @@ void apc_iterator_obj_init(apc_iterator_t *iterator, zval *search, zend_long for
 	iterator->key_idx = 0;
 	iterator->chunk_size = chunk_size == 0 ? APC_DEFAULT_CHUNK_SIZE : chunk_size;
 	iterator->stack = apc_stack_create(chunk_size);
```
madmajestro:
```suggestion
	iterator->stack = apc_stack_create(iterator->chunk_size);
```
I think it makes sense to init both stacks with the same capacity. In addition, this respects `APC_DEFAULT_CHUNK_SIZE` here, too.

apc_iterator.c
```diff
@@ -199,7 +203,8 @@ static int apc_iterator_check_expiry(apc_cache_t* cache, apc_cache_entry_t *entr
 static size_t apc_iterator_fetch_active(apc_iterator_t *iterator) {
 	apc_cache_t *cache = apc_user_cache;
 	size_t count = 0;
-	apc_iterator_item_t *item;
+	int i;
+	apc_stack_t *entries = iterator->entries;
```
madmajestro: I propose to remove this variable. We can use `iterator->entries` below.

apc_iterator.h
```diff
@@ -53,7 +53,8 @@ typedef struct _apc_iterator_t {
 							 /* fetch callback to fetch items from cache slots or lists */
 	size_t slot_idx;           /* index to the slot array or linked list */
 	size_t chunk_size;         /* number of entries to pull down per fetch */
-	apc_stack_t *stack;      /* stack of entries pulled from cache */
+	apc_stack_t *stack;      /* stack of items pulled from cache */
+	apc_stack_t *entries;    /* scratch stack of pinned entries, reused per fetch */
```
madmajestro:
```suggestion
	apc_stack_t *entries;    /* stack of pinned entries */
```

I propose to keep the comment similar for both stacks to prevent confusion (e.g, both are reused per fetch)

package.xml
madmajestro, line 68: Can you add a changelog entry?

madmajestro (commented):
I only noticed a few minor things / nits. I think we're almost there.

apc_iterator.c
iliaal: Removed, using iterator->entries directly in both fetch functions.

package.xml
iliaal: Added.

madmajestro (approved):
LGTM!

Thank you!

PR #629, opened 2026-07-24
Fix build against PHP 8.6 (php_verror params argument removed)

andypost:
php-src commit [0a12b3e8268](https://github.com/php/php-src/commit/0a12b3e826811fbebbcb80d2b959f7774d877b23) (php/php-src#22865, merged 2026-07-24) removed the docref `params` argument from `php_verror()`, changing its signature from 5 to 4 arguments. Since then APCu fails to compile against master:

```
/tmp/apcu/apc.c:44:17: error: too many arguments to function 'php_verror'; expected 4, have 5
   44 |                 php_verror(NULL, "", verbosity, format, args);
```

This adds a version-guarded `apc_php_verror()` wrapper macro (a `#if` cannot live inside the `APC_PRINT_FUNCTION` macro body), so the code builds against both PHP 8.6-dev and older versions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JNDQDyaiHPPMtsno9Kyo4t


andypost:
Faced it at Drupal's nightly pipeline https://git.drupalcode.org/project/drupalci_environments/-/jobs/11143560

apc.c
madmajestro, line 59: Generating functions with a macro-mechanism seems unnecessarily complex here. Especially since all the generated functions do pretty much exactly what `php_error_docref()` does. I know it was like that before, but I think it's time to change it, now.

It's probably better to remove these functions / macros completely from `apc.c` and replace the definitions in `apc.h` with macros that propagate the variadic data directly to `php_error_docref()`.

Therefore, i propose to replace this...
https://github.com/krakjoe/apcu/blob/bc9e296bc8d01ba774ba2ddd7b3f484217278d7d/apc.h#L80-L84

by something like this:
```
/* convenience macros for console output */
#define apc_error(...) php_error_docref(NULL, E_ERROR, __VA_ARGS__)
#define apc_warning(...) php_error_docref(NULL, E_WARNING, __VA_ARGS__)
#define apc_notice(...) php_error_docref(NULL, E_NOTICE, __VA_ARGS__)

#ifdef APC_DEBUG
# define apc_debug(...) php_error_docref(NULL, E_NOTICE, __VA_ARGS__)
#else
# define apc_debug(...) do {} while (0)
#endif
```

andypost: Sounds like great idea and clean-up but it breaks ABI so not clear if it allowed

andypost:
Opened #630 as an alternative that implements @madmajestro's [review suggestion](https://github.com/krakjoe/apcu/pull/629#pullrequestreview-4777474871): replace the generated console functions with macros forwarding to `php_error_docref()`. It fixes the same 8.6 build break and supersedes this PR. Trade-off (removes four `PHP_APCU_API` exported symbols) is written up there — committer's call which approach to take.

andypost:
After a closer look I recommend closing this PR in favor of #630.

Reviewing my own diff here turned up a structural defect: for `apc_error` the verbosity is `E_ERROR`, and `php_error_docref()` on that level ends in `zend_bailout()` — a longjmp that never returns (main/main.c, the E_ERROR case). So the `efree(buf)` after the call is unreachable on that path and the `vspprintf`-allocated buffer leaks. Harmless in practice (request-scoped emalloc, reclaimed at shutdown right after — the request is terminating anyway), but a debug-ZMM build or Valgrind run of the error path will flag it.

The defect can't be patched within this PR's shape: the buffer can't be freed before the call (it *is* the argument) nor after (bailout jumped away). The only fix is to not allocate — i.e. forward format+args directly to `php_error_docref()`, which is exactly what #630 does per @madmajestro's suggestion. Any repair of this PR converges on that one.

Both PRs are green across the full CI matrix including the 8.6 job. Leaving this open only as the minimal fallback if the `PHP_APCU_API` symbol removal in #630 is unwanted — otherwise #630 is the better change and this can be closed.

madmajestro:
I am closing this as it has been superseded by #630, which was recently merged.

PR #630, opened 2026-07-25
Simplify console output helpers

andypost:
## Simplify console output helpers

Follow-up to #629, per @madmajestro's review suggestion there.

Replaces the `APC_PRINT_FUNCTION` generator in `apc.c` — which produced four
exported `apc_error` / `apc_warning` / `apc_notice` / `apc_debug` functions —
with thin variadic macros in `apc.h` that forward straight to
`php_error_docref()`:

```c
#define apc_error(...)   php_error_docref(NULL, E_ERROR,   __VA_ARGS__)
#define apc_warning(...) php_error_docref(NULL, E_WARNING, __VA_ARGS__)
#define apc_notice(...)  php_error_docref(NULL, E_NOTICE,  __VA_ARGS__)
```

Net −16 lines, and it **supersedes #629**: the old generator body called
`php_verror()`, whose signature changed in php/php-src@0a12b3e8268 (params arg
removed) and broke the build against PHP 8.6-dev. Forwarding to
`php_error_docref()` sidesteps `php_verror()` entirely and builds on all
supported versions. If this lands, #629 can be closed.

### Trade-off: ABI

**Pro** — simpler (no `apc.c` code, no `va_list`/`vspprintf`/`efree`
plumbing), one fewer indirection, and format-string checking is preserved via
`php_error_docref`'s own `printf` attribute. `apc_debug` also stops evaluating
its arguments when `APC_DEBUG` is off (`do {} while (0)`), which is the usual,
slightly-better behaviour.

**Con** — the four functions were declared `PHP_APCU_API`, i.e. **exported
symbols**; turning them into macros removes them from the ABI. They're internal
console helpers (no in-tree or `bc/` caller takes their address or uses them
outside a direct call), so the practical risk is nil — but it is technically a
public-symbol removal, hence a separate PR for you to decide rather than folded
into the #629 build-fix.

### Verification

- All translation units that use these macros compile clean against php-src
  `master` (new 4-arg `php_verror` present in `main/php.h`).
- Counter-check: pristine `master` apcu fails the same compile on the
  `php_verror` argument mismatch — the exact break #629 addresses — confirming
  the fix is real and not a no-op.
- `APC_DEBUG=1` build compiles; no `-Wunused` fallout from the disabled
  `apc_debug`.


andypost:
Supersedes #629 (the minimal `php_verror` build-fix). This PR fixes the same PHP 8.6-dev build break and additionally applies @madmajestro's cleanup suggestion from that review. If this is preferred, #629 can be closed; if the ABI removal is unwanted, #629 remains the minimal fix.

andypost:
Pushed a small refinement: with `APC_DEBUG` off, `apc_debug` is now

```c
do { if (0) php_error_docref(NULL, E_NOTICE, __VA_ARGS__); } while (0)
```

instead of a bare `do {} while (0)`. The `if (0)` keeps compile-time `-Wformat` checking of debug call sites in normal (non-debug) builds at zero runtime cost — verified that a deliberately mismatched format in an `apc_debug()` call now warns even without `--enable-apcu-debug`. Previously a bad debug format would only surface in debug builds, which CI doesn't exercise.

madmajestro (approved):
LGTM! Thank you!

I think this ABI break is more of a theoretical issue. No other PHP extension should rely on APCu logging functions that were quite obviously intended for internal use. Especially since that functionality is also available via `php_error_docref()` in the PHP core. Not even the `apcu-bc` extension depends on these functions.

andypost:
Thank you, awaiting beta stability of 8.6 now

